PR #154, opened 2019-05-17
Unit tests added for cli commands

imsarath:
Unit testing started. Test cases added for BALSAMIC CLI. 

How to test:
 - Change the directory to BALSAMIC - `cd BALSAMIC`
 - Pull the code from `tests` branch
 - Run this `pytest tests/` 
 
Expected outcome:
 - It should run without error.
 - 15 test cases should be passed with two warnings.


tests/commands/config/test_config_sample.py
```diff
@@ -0,0 +1,41 @@
+import os
+import json
+import pytest
+
+from BALSAMIC.commands.config import get_config, write_json
+
+
+def iterdict(dic):
```
hassanfa: Maybe we can move it to `cli_utils.py` and add a unittest for it here instead.

tests/commands/config/test_config_sample.py
```diff
@@ -0,0 +1,41 @@
...
+    ref_json = json.load(open(config_files['reference'], 'r'))
+
+    tmp = tmp_path / "tmp"
+    tmp.mkdir()
+    output_json = tmp / "output.json"
+    write_json(ref_json, output_json)
+    output = output_json.read_text()
+
+    for key, value in iterdict(ref_json):
+        print(key)
```
hassanfa: Lets remove this.

PR #157, opened 2019-05-23
Added new testcases for config-sample and UMI flag

imsarath:
1. Functional testing added for config-sample command.
  2. To get analysis type, `--umi` flag was added and `--analysis_type` was removed from config-sample command.
  3. Based on this umi flag and normal sample, balsamic will calculate the analysis type.

**How to Test:**
 1. Change the directory to BALSAMIC - `cd BALSAMIC`
 2. Pull the code from `test_sample` branch
 3. Run this `pytest tests/` - for functional testing

For `--umi` implementation:
 1. Run `balsamic config sample --help`, to test implementation of `--umi` option
 2. change the directory - `tests/test_data`
 3. `--umi` option is a flag. Here, we are testing for `paired` analysis type. so, run balsamic with normal sample and without umi flag,
`balsamic config sample     --tumor fastq/S1_R_1.fastq.gz     --normal fastq/S2_R_1.fastq.gz     --panel-bed references/GRCh37/panel/panel.bed     --sample-id id1     --analysis-dir ./    --output-config id1_analysis.json     --reference-config references/reference.json`

**Expected outcome:**
 1. Testing should be run without error. 25 test cases should be passed.
 2. BALSAMIC should be configured without error while running the command for `paired` analysis.
 3. DAG file should be created.
 4. `-a / --analysis_type` option should be removed from `config sample` command

### DIFF
--- a/tests/commands/config/test_config_sample.py
+++ b/tests/commands/config/test_config_sample.py
@@ -128,3 +128,33 @@ def test_check_exist_error():
 
         #THEN It will raise the error 
         assert "FileIsNotFound" in str(error.value)
+
+def test_get_analysis_type():
+    #GIVEN umi flag(boolean value) and normal fq file 
+    umi_true = True 
+    normal_valid = 'normal.fastq'
+    umi_false = False 
+    normal_invalid = ''
+    #WHEN passing values 
+    #THEN it will return possible analysis type
+    assert 'paired_umi' == get_analysis_type(normal_valid, umi_true)
+    assert 'single_umi' == get_analysis_type(normal_invalid, umi_true)
+    assert 'paired' == get_analysis_type(normal_valid, umi_false)
+    assert 'single' == get_analysis_type(normal_invalid, umi_false)
+
+def test_get_output_config():
+    #GIVEN a config arg and sample id 
+    sample_id = 'test_sample'
+    config_json = 'config.json'
+    _config_json = ''
+
+    #WHEN passing values 
+    config = get_output_config(config_json, sample_id)
+    _config = get_output_config(_config_json, sample_id)
+    #THEN it will return config json file
+    assert config != ''
+    assert _config != ''
+    assert config.split('.')[-1] == 'json'
+    assert _config.split('.')[-1] == 'json'
+
+
--- a/tests/commands/config/test_config_sample.py
+++ b/tests/commands/config/test_config_sample.py
@@ -9,7 +9,6 @@ from BALSAMIC.commands.config import get_config, write_json, merge_json, \
 from BALSAMIC.tools import iterdict
 
 
-
 def test_get_config(config_files):
     # GIVEN the config files name
     files = [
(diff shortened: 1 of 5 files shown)


hassanfa:
👍 CLI option for single/pair with and without umi works. Well done.

I'll wait for your linting and style fix to review the code.

tests/commands/config/test_config_sample.py
```diff
@@ -1,8 +1,11 @@
 import os
```
hassanfa: I think `os` is not used.

tests/commands/config/test_config_sample.py
```diff
@@ -1,8 +1,11 @@
...
 import json
 import pytest
+from datetime import datetime
```
hassanfa: move above `import pytest`

tests/commands/config/test_config_sample.py
```diff
@@ -36,3 +39,145 @@ def test_write_json(tmp_path, config_files):
...
+        #GIVEN a invalid dict
+        ref_json = {"path": "/tmp", "reference": ""}
+        tmp = tmp_path / "tmp"
+        tmp.mkdir()
+        output_json = tmp / "/"
+
+        #WHEN passing a invalid dict
+        value = write_json(ref_json, output_json)
+
+    assert "Is a directory" in str(error.value)
```
hassanfa: Can you please add a comment for `#THEN`?

tests/commands/config/test_config_sample.py
```diff
@@ -36,3 +39,145 @@ def test_write_json(tmp_path, config_files):
...
+    #GIVEN a dict and json file
+    ref_dict = json.load(open(config_files['reference'], 'r'))
+
+    json_file = config_files['sample']
+
+    # WHEN passing dict and json file to merge
+    merge_dict = merge_json(ref_dict, json_file)
+
+    # THEN It will merge both the data and return dict
+    assert isinstance(merge_dict, dict) == True
```
hassanfa: No need for `== True`

tests/commands/config/test_config_sample.py
```diff
@@ -36,3 +39,145 @@ def test_write_json(tmp_path, config_files):
...
+    umi_true = True
+    normal_valid = 'normal.fastq'
+    umi_false = False
+    normal_invalid = ''
+    #WHEN passing values
+    #THEN it will return possible analysis type
+    assert 'paired_umi' == get_analysis_type(normal_valid, umi_true)
+    assert 'single_umi' == get_analysis_type(normal_invalid, umi_true)
+    assert 'paired' == get_analysis_type(normal_valid, umi_false)
+    assert 'single' == get_analysis_type(normal_invalid, umi_false)
```
hassanfa: In all four above asserts, swap the comparisons.

```suggestion
    assert get_analysis_type(normal_invalid, umi_false) == 'single'
```

tests/commands/config/test_config_sample.py
```diff
@@ -36,3 +39,145 @@ def test_write_json(tmp_path, config_files):
...
+    analysis_dir = './'
+    analysis_type = 'paired'
+
+    #WHEN passing args into the function
+    sample_config = get_sample_config(sample_config, sample_id, analysis_dir,
+                                      analysis_type)
+    date = datetime.now().strftime("%Y-%m-%d %H:%M")
+
+    #THEN it will return the updated python dict with given values
+    assert isinstance(sample_config, dict) == True
```
hassanfa: There is no need for `True`

tests/commands/config/test_config_sample.py
```diff
@@ -36,3 +39,145 @@ def test_write_json(tmp_path, config_files):
...
+    panel_bed = config_files["panel_bed_file"]
+
+    #WHEN passing args to that function
+    json_out = set_panel_bed(ref_json, panel_bed)
+
+    #THEN It will add two more items into the dict
+    assert "chrom" in json_out['bed']
+
+
+def test_set_panel_bed_error(config_files):
```
hassanfa:
```suggestion
def test_set_panel_bed_error(config_files):
    with pytest.raises(Exception, match=r"No such file or directory") as err:
        #GIVEN test reference json file and invalid panel bed file path
        ref_json = json.load(open(config_files['reference'], 'r'))

        panel_bed = "panel/panel.bed"

        #WHEN passing args to that function
        
        
        assert set_panel_bed(ref_json, panel_bed)
```

imsarath:
Added new test cases for config-sample and UMI flag

PR #160, opened 2019-05-28
Install balsamic through conda container

hassanfa:
This PR addresses issues with https://github.com/Clinical-Genomics/cg/issues/364 

To test this PR:
1. `git checkout install_singularity`
2. `./install.sh -s S -d 190528_HFA -p ${CONDA_ENVS_PATH} -c`

### DIFF
--- a/BALSAMIC/commands/install/install.py
+++ b/BALSAMIC/commands/install/install.py
@@ -195,7 +195,6 @@ def install(context, input_conda_yaml, env_dir_prefix, overwrite_env,
                         env_prefix,
                         fg='yellow'))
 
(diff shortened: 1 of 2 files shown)


hassanfa:
> Perfect. I have added a few enhancements for installation.
> 
> Enhancements:
> 
> 1. We can add a check for singularity installation if we use '-c' container option
> 2. If the installation stops at some point, It has to resume where it stops.

Valid points. Can you please open issues for these two? We can pick it when BALSAMIC is more stable.

imsarath:
Yeah sure, I will open an issue and add these two.

PR #187, opened 2019-07-11
New reference config format to all snakemake rules

imsarath:
This PR contains,
  * New reference config updated in all snakemake rules.
  * Made reference-config as a required option for `config sample` command

**How to Run GenerateRef**
 *  This command will create output reference dir and config.json within that dir.
  `$ balsamic config reference -c cosmic-key -o reference`
 * To run GenerateRef workflow, 
   `$ balsamic run reference -c reference/config.json --snakemake-opt --cores`
 * As generate-ref workflow is running on local mode, '--cores' option will use available cores in that machine.
  * It will generate all required reference files for BALSAMIC and `reference.json`
  * sample reference json format added to config directory. (https://github.com/Clinical-Genomics/BALSAMIC/blob/reference_json/BALSAMIC/config/reference_new.json)

 Note: Need to add `exon_bed` in GenerateRef workflow, It was replaced with the panel bed file as a placeholder in samba rule.


hassanfa:
Can you add details of how to run this new additions?

e.g.
1. `balsamic config reference`
2. `balsamic run reference`

hassanfa:
QC, single and paired mode, and reference creation passed.

PR #190, opened 2019-07-16
Refactored link fastq and added snakemake constructor in config-sample

imsarath:
**This PR contains,**
- Refactoring link fastq function
- Invoked Snakemake class into config-sample 
- Test cases added for newly written functions

**How to Test:**
- clone this repo, then `cd BALSAMIC`
- Run pytest `pytest tests/`

**Expected outcome:**
- Test cases should be passed without error


BALSAMIC/commands/config/sample.py
```diff
@@ -97,167 +98,103 @@ def get_sample_config(sample_config, sample_id, analysis_dir, analysis_type):
...
     """
-    Links fastq files inside the analysis directory
+    Creating fastq symlinks in given destination
     """
+    for src_file in src_files:
+        basename = os.path.basename(src_file)
+        des_file = os.path.join(des_path, basename)
+        try:
+            os.symlink(src_file, des_file)
+        except FileExistsError as error:
```
hassanfa: This shouldn't raise error. A warning stated that link exists and wont be created is enough.

BALSAMIC/commands/config/sample.py
```diff
@@ -283,73 +220,11 @@ def sample(
...
-            link_fastq(
-                os.path.abspath(normal_path),
-                os.path.abspath(fastq_path),
-                normal,
-                read_prefix,
-                check_fastq,
-                fastq_prefix,
-            )
+    # create dir for fastq symlink creation
+    fq_path = os.path.join(output_dir, 'fastq')
```
hassanfa: Can we change fq_path to `sample_id/analysis/fastq`? It's a long due change that I didn't want to make.

BALSAMIC/commands/config/sample.py
imsarath: It was written for testing purpose. I will remove that error.

imsarath:
Changes:
- Testcases for the snakemake workflow with dry-run option is added in this PR. 
- Refactored `tests/test_data` directory

PR #192, opened 2019-07-19
Added validation for link fastq and test cases

imsarath:
**This PR contains,**
- refactoring of link_fastq with try and exception
- Added test cases for new functions 

**How to Test:**
- clone this repo, then cd BALSAMIC
- Run pytest, `pytest tests/`

**Expected outcome:**
-  Test cases should be passed without error

### DIFF
--- a/BALSAMIC/commands/config/sample.py
+++ b/BALSAMIC/commands/config/sample.py
@@ -126,6 +126,22 @@ def link_fastq(src_files, des_path):
             )
 
 
+def get_fastq_path(file, fq_pattern):
+    # check the fastq if exists
+    file = os.path.abspath(file)
+    if Path(file).exists():
+        file_basename = os.path.basename(file)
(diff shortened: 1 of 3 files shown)


BALSAMIC/commands/config/sample.py
```diff
@@ -238,8 +243,7 @@ def configure_fastq(fq_path, tumor, normal, fastq_prefix):
 @click.pass_context
 def sample(context, umi, install_config, sample_config, reference_config,
            panel_bed, output_config, normal, tumor, sample_id, analysis_dir,
-           fastq_path, check_fastq, overwrite_config, create_dir,
-           fastq_prefix):
+           check_fastq, overwrite_config, create_dir, fastq_prefix):
```
hassanfa: `check_fastq` is not used anymore. imho, we can safely remove it.

PR #218, opened 2019-09-27
Sentieon T/N and Tumor only pipeline updated 

imsarath:
### This PR:

- TNscope and DNAscope added to sentieon rules
- TNscope filtering based on a machine learning model added
- DNAscope filtering disabled due to machine learning model issues
- Benchmark dir added to capture the processing time and CPU memory usage.
- Fixed fastp rule to write output json and html file for each sample.

### How to test:

### Expected outcome:

### Review:
- [x] Code review
- [x] Tests pass

### DIFF
--- a/BALSAMIC/commands/run/analysis.py
+++ b/BALSAMIC/commands/run/analysis.py
@@ -121,13 +121,15 @@ def analysis(context, snake_file, sample_config, run_mode, cluster_config,
             if files:
                 logpath = createDir(logpath, [])
                 scriptpath = createDir(scriptpath, [])
+                benchmarkpath = createDir(benchmarkpath, [])
 
     # Create result directory
     os.makedirs(resultpath, exist_ok=True)
-    os.makedirs(benchmarkpath, exist_ok=True)
+
     if not os.path.exists(logpath):
         os.makedirs(logpath, exist_ok=True)
         os.makedirs(scriptpath, exist_ok=True)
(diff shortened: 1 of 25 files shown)


hassanfa:
Looks good. I'll review it this weekend. A quick comment to add downloading ML models either to reference or to assets.

imsarath:
Now I will add ML models to assets directory, Once we fix the singularity PR, we will move this into the reference workflow flow (Just to avoid conflicts).

BALSAMIC/commands/run/analysis.py
```diff
@@ -123,6 +124,7 @@ def analysis(context, snake_file, sample_config, run_mode, cluster_config,
 
     # Create result directory
     os.makedirs(resultpath, exist_ok=True)
+    os.makedirs(benchmarkpath, exist_ok=True)
```
hassanfa: Move this under `if not...` in next line. It shouldn't create benchmark dir if in dry run mode.

BALSAMIC/commands/run/analysis.py
```diff
@@ -111,6 +111,7 @@ def analysis(context, snake_file, sample_config, run_mode, cluster_config,
     logpath = sample_config['analysis']['log']
     scriptpath = sample_config['analysis']['script']
     resultpath = sample_config['analysis']['result']
+    benchmarkpath = sample_config['analysis']['benchmark']
```
hassanfa: Move this after line 121 `if files:`

Issue #229, opened 2019-10-12
SLURM specific details for WES runs


hassanfa:
Similar to "SLURM specific details for WGS runs" #220

Overall non-alignment rules and those that are not related to mutect2 should be reduced.

Use the results below for adjusting required time:

![image](https://user-images.githubusercontent.com/1086877/66703103-44701800-ed0f-11e9-8c2d-89249e663072.png)

@imsarath when you're back, can you please take care of this? The same way as you did for WGS rules.

hassanfa:
Addressed in #235

PR #235, opened 2019-10-23
SLURM specific details updated for WES analysis

imsarath:
### This PR:

- Updated cluster config file with a time limit and threads for all rules (SLURM)
   - Addressed #229 
- Benchmark added for all variant callers
- Simplified merge bam rules

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] New code is executed and covered by tests
- [ ] Tests pass

### DIFF
--- a/BALSAMIC/config/cluster.json
+++ b/BALSAMIC/config/cluster.json
@@ -54,7 +54,7 @@
         "n": 16
     },
     "haplotypecaller_merge": {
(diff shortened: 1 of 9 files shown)


hassanfa:
👍 Looks good. I'm testing Mutect2 with `-nt` option once more, and see if we can adopt to this new more adjusted resources allocations.

PR #241, opened 2019-10-31
Feature: support for qsub

hassanfa:
### This PR:

- Adds support for pbs qsub

### How to test:

### Expected outcome:

### Review:
- [x] Code review
- [x] Tests pass

This changes in this PR are incompatible with previous releases. Thus it will be a major version


hassanfa:
@imsarath Good work on this. I made two small changes to make the dependency tree work. I am making this PR on your behalf. These items needs to be done:
1. Rename `sbatch.py` to `scheduler.py`
2. Adds tests for `QsubScheduler` and `SbatchScheduler` class.
3. Mock subprocess to make it work for `sbatch` and `qsub`.

imsarath:
Thanks, @hassanfa. I will update the test cases in this PR

hassanfa:
The tests are already there for the two classes. 😆 Only a test for `scheduler.py` argparse is needed. But let's do that in another PR.

PR #244, opened 2019-11-05
Sentieon callers for WES

imsarath:
### This PR:

- Sentieon variant callers (TNhaplotyper - Somatic, DNAscope - Germline) added for WES or Targeted panel workflow

### How to test:

### Expected outcome:

### Review:
- [x] Code review
- [x] Tests pass


imsarath:
Thanks, @hassanfa. Since we are not using sentieon's machine learning models in WES workflow, I just removed ML env variables from workflow and As you suggest we are controlling the sentieon variant callers using sentieon variable.

PR #248, opened 2019-11-07
Refactor benchmark files

hassanfa:
### This PR:

- Refactors benchmark filenames with snakemake rule names due to some output files having the same _filename_ but with different path (e.g. vep and vcf path)

### How to test:

### Expected outcome:

### Review:
- [x] Code review
- [x] Tests pass


hassanfa:
> Minor observation, snakemake rules have only 2 space indentation. 👍

😄  True that, reformatting that many rule was a gruesome task and maybe for another PR.

PR #316, opened 2020-05-13
Make fastp explicitly cuts adapters for PE

hassanfa:
### This PR:

- See title

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [x] Tests pass


hassanfa:
👍 Analysis finished without any error:

image

hassanfa:
NOTE: not ready to merge. Testing again. I think we should split quality trim, umi trim, and adapter trimming. It seems fastp's order might affect templates with short length, which will eventually affect downstream analysis.

If we have `adapter --- umi --- template` in each read, according to fastp order, umi will be trimmed first ON the `adapter` part, and actual umi will be NOT be trimmed.

I'll split up parameters and test again.

PR #328, opened 2020-06-08
Add libiconv dependency to BALSAMIC_py36

### LINKED ISSUE #324: Possible bug PICARD and Vardict fail on linked R libraries
**Describe the bug**
VarDict and PICARD use R internally to filter and/or prepare plots. They fail on missing R library and in particular: `lib/R/bin/exec/R: error while loading shared libraries: libiconv.so.2 cannot open shared object file: No such file or directory` 

**To Reproduce**
use Balsamic container `latest`

**If workflow, which rules**
All vardict variant calling rules and all picard metric rules

**Desktop (please complete the following information):**
`4.2.4` output

**Additional context**
Probably linked to new dependencies for various conda packages.

hassanfa:
### This PR:

- Adds libiconv to solve fix #324 
- Fix conda commandline bug with correct option `-all` to `--all`
- Add ARG to Dockerfile to make it possible to pull from a different branch

### How to test:

### Expected outcome:

### Review:
- [ ] Code review

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,4 +2,5 @@
 # the repo. Unless a later match takes precedence,	
 # @global-owner1 and @global-owner2 will be requested for	
 # review when someone opens a pull request.	
-*       @hassanfa @ashwini06
+* @hassanfa @ashwini06
+*.py @mropat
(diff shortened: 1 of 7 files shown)


hassanfa:
@ashwini06 this is a small one

Some background: We have fixed version of each bioinfo tool, but it seems it is not the way to go. Apparently libiconv is not available anymore while building the environment. I don't know when this happened, but adding libiconv to to explicitly install it seems to work.

hassanfa:
I am merging this, since it is just a fix for missing dependency. 
image

PR #330, opened 2020-06-08
add maria as codeowner for all .py files

hassanfa:
This PR adds Maria as codeowner for all *py files.


hassanfa:
I merged this as it is not an actual code.

PR #344, opened 2020-07-10
Add normal sample to run_validate

hassanfa:
### This PR:

- Adds normal sample to run_validate script
- Removes run_validate from .travis.yaml

### How to test:

### Expected outcome:

### Review:
- [ ] Code review

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,11 +13,6 @@ install:
 
 script:
   - py.test --cov=BALSAMIC -rsxv tests/* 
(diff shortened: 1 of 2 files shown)


Mropat:
Could you clarify:
- Will the change in run_validate.sh alter the way config case command is used?
- If run_validate is removed from Travis, how will it be used in the future?

hassanfa:
- no, it will not alter it. This is actually fixing it to properly run config case with a normal sample.
- run_validate won't work on travis. Travis doesn't have singularity available. It'll just a wrapper script to help validate new releases.

PR #352, opened 2020-07-15
Refactor/delivery workflow

hassanfa:
## This PR:

- Refactors delivery part of workflow
- Add report capability to Alignment workflow
- suppress report warnings from PR #348 

### How to test:

- Automatic and continuous test pass

### Expected outcome:
- Installation, unit and integration tests pass

### Review:
- [ ] Code review
- [ ] New code is executed and covered by tests, and test approve

PR #354, opened 2020-07-16
fix, workdir as path

hassanfa:
### This PR:

- Work directory as path in delivery report

### How to test:

### Expected outcome:

### Review:
- [ ] Code review


hassanfa:
forgot to commit on previous PR #352. since we have it as proper posix path in run_analysis, should have it here as well.

PR #357, opened 2020-07-17
Panel bind path fix

Mropat:
### This PR:

Fixed: Bind path is not added as None if analysis type is wgs

### Review and tests: 
- [ ] Tests pass
- [ ] Code review

### DIFF
--- a/BALSAMIC/commands/config/case.py
+++ b/BALSAMIC/commands/config/case.py
@@ -94,7 +94,16 @@ def case_config(context, case_id, umi, umi_trim_length, adapter_trim,
         )
         raise click.Abort()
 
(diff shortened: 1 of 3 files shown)


Mropat:
Ready for review @hassanfa

hassanfa:
This solution won't work properly. WGS config should not have any `panel` info int he config. The config case model seems to be faulty.

```
    "panel": {
        "capture_kit": null,
        "chrom": null
    }
```

Above has to be removed from config file.

hassanfa:
I changed bioinfo_tools to optional, and excluded none values from pydantic. Let's remember to always include None values, as they can break code.

Also, I prefer explicit `append` or `extend` over `+`,  it makes clear what is being done and not a guess game.

PR #358, opened 2020-07-20
Changes panel bed model

Mropat:
### This PR:

Refactor: Move all optional models into Balsamic config model.

### Review and tests: 
- [ ] Tests pass
- [ ] Code review

Motivation:
Enabling flag "exclude_none"=True during model construction, solves the panel bed bug. 
This suggestion would solve the previous problem in desired way with minimum changes to the structure of the code.


hassanfa:
coverage dropped by 0.005%... Not terrible, not great 
image

PR #371, opened 2020-07-22
Refactor reference generate workflow

hassanfa:
### This PR:

Added: Added possibility to extend workflow with newer genome version
Changed:
Reference workflow is now based on various reference types and model
`genome/refseq.flat` renamed to `genome/refGene.flat`

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


hassanfa:
Ready to review!

BALSAMIC/workflows/GenerateRef
```diff
@@ -154,97 +143,80 @@ rule all:
...
+            log_file = output_file + ".log"
+
+            if ref.url.scheme == "gs":
+                cmd = "gsutil cp -L {}.log {} -".format(log_file, ref.url)
             else:
-                shell("wget -a {ref[1]}.log -O {ref[1]} {ref[0]}")
+                cmd = "wget -a {}.log -O - {}".format(log_file, ref.url)
+
+            if ref.secret:
+                response = requests.get(ref.url, headers={'Authorization': 'Basic %s' % ref.secret })
```
Mropat, line 165: The requests.get() command can often fail, due to timeouts, network errors, or for no good reason whatsoever. 
Could we maybe catch this with a retry clause, and if all fails, raise an error to handle later in CLI?

hassanfa:
tested workflow and reference files are same between this PR and previous reference file. Here is MD5 checks:
image

hassanfa:
More MD5 evidence: 
image

BALSAMIC/workflows/GenerateRef
```diff
@@ -99,29 +88,29 @@ singularity_image = config['singularity']['image']
...
+        known_indel_1kg = known_indel_1kg_url.get_output_file + ".gz",
+        cosmic_vcf = cosmicdb_url.get_output_file + ".gz",
+        variants_idx = expand( os.path.join(vcf_dir,"{vcf}.gz.tbi"), vcf=VCF),
+        vep = directory(vep_dir),
+        wgs_calling = wgs_calling_url.get_output_file,
+        genome_chrom_size = genome_chrom_size_url.get_output_file 
     output:
         finished = os.path.join(basedir,"reference.finished"),
-        reference_json = reference_json,
+        reference_json = os.path.join(basedir, "reference.json"),
```
ashwini06, line 110: maybe, something like this 


`reference_json = join(basedir, "reference.json"),`

BALSAMIC/workflows/GenerateRef
```diff
@@ -4,71 +4,60 @@
 
 import os
 import hashlib
+
 from datetime import date
```
ashwini06, line 9: from os.path import join

tests/utils/test_models.py
```diff
@@ -1,15 +1,168 @@
...
+        "url": "gs://domain/file_name",
+        "file_type": "fasta",
+        "gzip": True,
+        "genome_version": "wrong_genome",
+        "output_file": dummy_output_file,
+        "output_path": dummy_output_path,
+    }
+
+    #WHEN building the model
+
```
ashwini06, line 161: Missing condition for #WHEN??

BALSAMIC/utils/constants.py
```diff
@@ -26,51 +23,200 @@
...
+    "analysis_type": "tumor_only",
+    "description": "General purpose filters used for filtering VarDict"
+}
+
+#reference related constants
+VALID_REF_FORMAT = ["fasta", "vcf", "text", "gtf", "gff"]
+VALID_GENOME_VER = ["hg19", "hg38"]
+
+#reference files
+REFERENCE_FILES = {
```
ashwini06, line 122: Missing reference files for hg38?

BALSAMIC/utils/models.py
```diff
@@ -336,3 +346,111 @@ def abspath_as_str(cls, value):
...
+        return value
+
+    @property
+    def get_output_file(self):
+        """return output file full path"""
+        output_file_path = Path(self.output_path, self.output_file).as_posix()
+        return output_file_path
+
+    @property
+    def write_md5(self):
```
ashwini06, line 394: Codefactor showing warning for this function name. Please check it

BALSAMIC/workflows/GenerateRef
hassanfa: That's a good point. I'll add a try..except for `response`.

BALSAMIC/workflows/GenerateRef
ashwini06: Just a suggestion for all the `os.path.join` lines.

BALSAMIC/workflows/GenerateRef
hassanfa: Good point. I'll add this.

tests/utils/test_models.py
hassanfa: Good catch! when/then are badly placed.

BALSAMIC/utils/constants.py
hassanfa: this is a placeholder. hg38 will follow in another PR.

BALSAMIC/utils/models.py
hassanfa: Yes, i saw that. We are just checking for integrity of the files and security warning here is for safety of "MD5" encryption. since we are not doing any encryption, I think we can safely ignore this.

The goal is to check integrity of MD5 files while creating config, to make sure what we have in config is same as original files.

hassanfa:
Codefactor complaint is a bit strange. MD5 safety is not a concern for now. Merging this.

PR #376, opened 2020-07-26
Refactor/delivery file

hassanfa:
### This PR:

Changed:
- delivery file format simplified
- added suffix: all and pass to VEP's rule output.

Added:
- delivery option to CLI and what to delivery
- delivery params in rules that are needed to be delivered: fastp, multiqc, vep_somatic, ngs_filter_vardict, cnvkit_paired, cnvkit_single, mergeBam_normal, mergeBam_tumor (tumor_only and tumor_normal)

### Review and tests: 
- [x] Tests pass
- [x] Code review


hassanfa:
Delivery file created for T and TN panel: 
image

Delivery file for T and TN WGS:
image

hassanfa:
Ability to deliver new items through CLI options **in addition** to what it can deliver. Here BAM/CRAM files are delivered, in addition to default ones:
image

Ability to deliver new items  through CLI options and **only** delivering that. Here only fastp and multiqc files are delivered:
image

BALSAMIC/commands/report/deliver.py
```diff
@@ -64,174 +95,120 @@ def deliver(context, sample_config, analysis_type):
     yaml_write_directory = os.path.join(result_dir, "delivery_report")
```
Mropat, line 95: Small comment, regarding dependencies.
Would it be possible to use pathlib instead of os.path? Functionally, it doesn't make a big difference, but it would be easier to stick to one import for handling paths, whichever we choose

BALSAMIC/snakemake_rules/annotation/varcaller_filter.rule
```diff
@@ -11,11 +11,12 @@ VARDICT = VarCallerFilter.parse_obj(VARDICT_SETTINGS)
...
   input:
-    vcf = vep_dir + "{var_type}.somatic.{case_name}.vardict.vcf.gz",
+    vcf = vep_dir + "{var_type}.somatic.{case_name}.vardict.all.vcf.gz",
   output:
-    vcf_filtered = vep_dir + "{var_type}.somatic.{case_name}.vardict.filtered.vcf.gz",
-    vcf_pass = vep_dir + "{var_type}.somatic.{case_name}.vardict.filtered.pass.vcf.gz",
+    vcf_filtered = vep_dir + "{var_type}.somatic.{case_name}.vardict.all.filtered.vcf.gz",
+    vcf_pass = vep_dir + "{var_type}.somatic.{case_name}.vardict.all.filtered.pass.vcf.gz",
   params:
+    housekeeper_id = {"id": ["case"], "tags": ["filtered_vcf", "scout"]},
```
Mropat, line 19: Would it be possible to use dash-separated tag names as housekeeper tags? I imagine it does not make a big difference functionally. Dash-separating the tags would however follow same naming convention as MIP are using for hk tags. 
At the end of the day, it is up to the end user whichever they prefer.

BALSAMIC/commands/report/deliver.py
hassanfa: I prefer pathlib. I think we agreed to use pathlib, didn't we?

I'll create an issue to migrate all os.path to pathlib.Path. It needs a bit of careful planning as there is ancient code in Snakemake rules that might not like it.

Would you accept if we postpone this refactoring for after vacation times, please?

BALSAMIC/snakemake_rules/annotation/varcaller_filter.rule
hassanfa: absolutely! I'll fix them.

BALSAMIC/commands/report/deliver.py
hassanfa: issue created #377

PR #387, opened 2020-08-17
Fix ambiguous delivery wildcard for vep_somatic

hassanfa:
### This PR:

Fixed: ambiguous delivery wildcard files

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


hassanfa:
@Mropat could you try this on case: `bosssponge`  please? I guess we can use `S_BALSAMIC`.

Issue #370, opened 2020-07-22
[ Failed Analysis ]BALSAMIC.daringmidge.mutect2_tumor_normal


ashwini06:
**Case or sample information**
Internal sample id or case id: daringmidge
Amount read-pairs: tumor 93 Mreads; normal 124 Mreads
Sequence type (WES, TGA, WGS): TGA GMCKsolid

**Error message**
`Slurm Job_id=521003 Name=BALSAMIC.daringmidge.mutect2_tumor_normal.120.sh Failed, Run time 14:00:13, TIMEOUT, ExitCode 0`

**Link to the error message**
`/home/proj/production/cancer/cases/daringmidge/logs/BALSAMIC.daringmidge.mutect2_tumor_normal.120.sh_521003.err`

**Reason for failure**
`slurmstepd: error: *** JOB 521003 ON compute-0-9 CANCELLED AT 2020-07-21T15:53:12 DUE TO TIME LIMIT ***`

**If workflow, which rules**
`snakemake_rules/variant_calling/somatic_tumor_normal.rule`

**Version (please complete the following information):** 
`4.3.0`

**Additional context**
Possible Solution:
According to err log, mutect2 ran for 14 hours and exited due to time limit
Rerun by increasing compute time for daringmidge tumor normal to 15 hours instead

hassanfa:
Fixed in #440

Issue #367, opened 2020-07-21
[ Failed Analysis ] BALSAMIC.meetcod.strelka_tumor_normal


hassanfa:
**Case or sample information**
Internal sample id or case id: meetcod
Amount read-pairs: normal 145M tumor 112M
Sequence type (WES, TGA, WGS): TGA gmcksolid

**Error message**
> Slurm Job_id=520270 Name=BALSAMIC.meetcod.strelka_tumor_normal.41.sh Failed, Run time 06:00:22, TIMEOUT, ExitCode 0

**Link to the error message**
`/home/proj/production/cancer/cases/meetcod/logs/BALSAMIC.meetcod.strelka_tumor_normal.41.sh_520270.err`

**Reason for failure**
Time out.

**If workflow, which rules**
Variant calling paired analysis. Strelka tumor-normal

**Version (please complete the following information):**
`4.3.0`

**Additional context**
Possible solution: increase analysis time to 12hours.

ashwini06:
Rerun analysis failed again 

`Slurm Job_id=523990 Name=BALSAMIC.meetcod.strelka_tumor_normal.41.sh Failed, Run time 12:00:06, TIMEOUT, ExitCode 0`

```slurmstepd: error: *** JOB 523990 ON compute-0-0 CANCELLED AT 2020-07-23T02:00:26 DUE TO TIME LIMIT ***```

hassanfa:
Increased time to 18hours and 18 threads, up from 12 hours and 12 threads. This is worrying me. Either Strelka fails to use multithreading options, or there are tough to call regions.

@ashwini06 Please restart the once again. Thanks!

ashwini06:
Thanks @hassanfa . Restarted the analysis now.

ashwini06:
Failed again

`Slurm Job_id=525656 Name=BALSAMIC.meetcod.strelka_tumor_normal.41.sh Failed, Run time 18:00:28, TIMEOUT, ExitCode 0`

`slurmstepd: error: *** JOB 525656 ON compute-0-4 CANCELLED AT 2020-07-24T06:42:37 DUE TO TIME LIMIT ***`

hassanfa:
addressed in in PR

Issue #366, opened 2020-07-21
[ Failed Analysis ] BALSAMIC.superbkite.strelka_tumor_normal


hassanfa:
**Case or sample information**
Internal sample id or case id: superbkite
Amount read-pairs: normal 270M and tumor 90M 
Sequence type (WES, TGA, WGS): TGA, gmcksolid

**Error message**
> Slurm Job_id=520624 Name=BALSAMIC.superbkite.strelka_tumor_normal.41.sh Failed, Run time 06:00:17, TIMEOUT, ExitCode 0

**Link to the error message**
`/home/proj/production/cancer/cases/superbkite/logs/BALSAMIC.superbkite.strelka_tumor_normal.41.sh_520624.err`

**Reason for failure**
Time out. Current time for Strelka tumor normal is 6hours, increased to 12hours.

**If workflow, which rules**
paired analysis, tumor normal variant calling

**Version (please complete the following information):**
`4.3.0`

**Additional context**
Possible solution:
Increase compute time for Strelka tumor normal to 12 hours instead.

hassanfa:
already fixed.

PR #392, opened 2020-08-24
increase analsyis run time for various tools

hassanfa:
### This PR:

Fixed: failed analysis run time for #370 #367 #366

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


Mropat (approved):
+1

hassanfa:
Proper way would be to scale these values based on amount of input reads. For now, manual updates are the only solution.

PR #431, opened 2020-09-29
Features: Adds qc rules to UMI workflow

ashwini06:
### This PR 
Corresponds to issue 2 in [cg-cancer-umi ](https://github.com/Clinical-Genomics/cgp-cancer-umi/issues/2)

**Added:** 

1.  Rules to calculate QC measures for UMI workflow. Implementation includes:
       
     - [x] Picard UmiAware
       
     - [x] Picard HSmetrics

     - [x] Noise Allele frequency       

     - [x]  Density plots for Allelic frequency distribution -  [test_output.pdf](https://github.com/Clinical-Genomics/BALSAMIC/files/5311985/test_output.pdf)

2. Thread parameters for the UMI rules in cluster.json file

**Changed:** 
Added wildcard {step} to variant calling and vep rules 
Initially, UMIextract bwa_alignment outputs the` sam` file. Now changed to `bam`

**Execution:**
`balsamic run analysis --sample-config demo/fairsalmon2/fairsalmon2.json --run-mode local --snakemake-opt "--cores 8" -a umi --run-analysis`

Screenshot 2020-09-29 at 12 30 57

**QC outputs**
Screenshot 2020-10-02 at 09 49 16

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


BALSAMIC/commands/config/case.py
```diff
@@ -110,7 +110,9 @@ def case_config(context, case_id, umi, umi_trim_length, adapter_trim,
             "analysis_type": "paired" if normal else "single",
             "sequencing_type": "targeted" if panel_bed else "wgs",
         },
-        reference=reference_dict,
+        SENTIEON_INSTALL_DIR="/home/proj/bin/sentieon/sentieon-genomics-201911/",
+	SENTIEON_LICENSE="10.10.10.1:8990",
```
hassanfa, line 114: Could we get these from env variables? What will happen if environment variables don't exist? Otherwise, set it in the workflow instead. Similar to other workflows.

BALSAMIC/snakemake_rules/umi/qc_umi.rule
```diff
@@ -0,0 +1,121 @@
...
+
+## CollectHSmetrics - median target coverage-required
+rule CollectHsMetrics_umi:
+    input:
+        fadict = (config["reference"]["reference_genome"]).replace(".fasta",".dict"),
+        bed = config["panel"]["capture_kit"],
+        bam = umi_dir + '{sample}.{step}.bam',
+        fa = config["reference"]["reference_genome"]
+    output:
+        mrkdup = qc_dir + '{sample}.{step}.collect_hsmetric'
```
hassanfa, line 43:
```suggestion
        mrkdup = qc_dir + '{sample}.{step}.collect_hsmetric_umi'
```

BALSAMIC/snakemake_rules/umi/qc_umi.rule
```diff
@@ -0,0 +1,121 @@
...
+    params:
+        conda = get_conda_env(config["conda_env_yaml"],"picard"),
+        baitsetname = os.path.basename(config["panel"]["capture_kit"]),
+        sample_id = '{sample}_{step}'
+    threads: get_threads(cluster_config, 'CollectHsMetrics')
+    singularity: singularity_image
+    benchmark:
+        benchmark_dir + "{sample}.{step}.collect_hsmetrics.tsv"
```
hassanfa, line 51:
```suggestion
        benchmark_dir + "{sample}.{step}.collect_hsmetrics_umi.tsv"
```

BALSAMIC/snakemake_rules/umi/qc_umi.rule
```diff
@@ -0,0 +1,121 @@
...
+        get_threads(cluster_config, 'mean_family_depth')
+    message:
+        "Calculating mean family depth for {params.sample_id}"
+    shell:
+        "source activate {params.conda}\n"
+        "samtools view {input.bam} | grep 'RX:Z:' | sed 's/.*RX:Z:\([ACGT-].*\).*/\\1/' | cut -f1 | grep -v 'N' | sort | uniq -c | sed -e 's/  */\\t/g' | cut -f2,3 > {output.temp_fl}\n"
+        "awk -F'\\t' '{{sum+=$1;}} END{{printf(\"{params.sample_id}_meandepth: \"sum/NR)}}' {output.temp_fl} > {output.totalsum}\n"
+
+## Noise AF = TotalAF_with_UMI/TotalAF_without_UMI - considering TNscope values
+rule noise_AF:
```
hassanfa, line 82: Interesting approach, I like it! Lets see how it will perform.

BALSAMIC/utils/models.py
```diff
@@ -332,7 +332,9 @@ class BalsamicConfigModel(BaseModel):
         reference : Field(Dict); dictionary containign paths to reference genome files
         panel : Field(PanelModel(optional)); variables relevant to PANEL BED if capture kit is used
         bioinfo_tools : Field(BioinfoToolsModel); dictionary of bioinformatics software and their versions used for the analysis
-        singularity : Field(Path); path to singularity container of BALSAMIC
+        SENTIEON_INSTALL_DIR:Field(str); Prefix to sentieon installation directory
+        SENTIEON_LICENSE:Field(str); sentieon license number
```
hassanfa, line 336: UMI changing often, maybe we can remove these and keep them in workflow for now? If you add them here, it might force your hand into fixing things that are not related to UMI.

It can be set on runtime in workflow. For example see: https://github.com/Clinical-Genomics/BALSAMIC/blob/master/BALSAMIC/workflows/VariantCalling_sentieon.smk line35-43.

BALSAMIC/utils/umiplots.py
```diff
@@ -0,0 +1,17 @@
...
+    umi.columns = ['id', 'AF']
+    umi['method'] = 'umiextract'
+    con = pd.read_csv(input_file2, sep='\t', header=None)
+    con.columns = ['id', 'AF']
+    con['method'] = 'consensuscall'
+    sns.kdeplot(umi['AF'], color='r', shade=True, Label=set(umi['method']))
+    sns.kdeplot(con['AF'], color='b', shade=True, Label=set(con['method']))
+    plt.xlabel('Allelic Frequency (AF)')
+    plt.ylabel('Probability Density')
+    return (plt.savefig(output_file))
```
hassanfa, line 17: You can move the whole thing into the rule. Otherwise, write a test for it 😛

BALSAMIC/utils/umiplots.py
```diff
@@ -0,0 +1,17 @@
+import sys
```
hassanfa, line 1: Seems like `sys` is not used.

BALSAMIC/utils/models.py
hassanfa: Also, i see that they are already set in workflow. If you've changed this due to my comment, I apologize. I might not have noticed it. `UMIWorkflow.smk` already has it set.

hassanfa (changes requested):
👍🏽 I know it is a draft, but I couldn't resist ;-)

hassanfa:
I changed the base branch to develop from master.

BALSAMIC/snakemake_rules/umi/qc_umi.rule
ashwini06: Not the best approach, I worry about the computation time. If I come up with an alternative solution I will change it :)

ashwini06:
@hassanfa Thank you for your comments on draft. I fixed all those, can you please have a look again.

hassanfa (approved):
Looks good!

hassanfa:
⭐  Merging it!

PR #433, opened 2020-10-02
Refactor/disable multiple variant callers

hassanfa:
### This PR:

Changed:
- Disable multiple variant callers via comma separation

### Review and tests: 
- [ ] Tests pass
- [ ] Code review

### DIFF
--- a/BALSAMIC/commands/run/analysis.py
+++ b/BALSAMIC/commands/run/analysis.py
@@ -97,7 +97,7 @@ LOG = logging.getLogger(__name__)
     '--disable-variant-caller',
     help=
     f'Run workflow with selected variant caller(s) disable. Use comma to remove multiple variant callers. Valid '
-    f'valules are: {list(VCF_DICT.keys())}',
+    f'values are: {list(VCF_DICT.keys())}',
 )
 @click.pass_context
 def analysis(context, snake_file, sample_config, run_mode, cluster_config,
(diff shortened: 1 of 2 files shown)


BALSAMIC/commands/run/analysis.py
```diff
@@ -93,9 +93,12 @@
     help='cluster mail type to send out email. \
               This will be applied to all jobs and override snakemake settings.'
 )
-@click.option('--disable-variant-caller',
-              type=click.Choice(list(VCF_DICT.keys())),
-              help='Run workflow with selected variant caller disable.')
+@click.option(
+    '--disable-variant-caller',
+    help=
```
ashwini06, line 98: dont you want to have `type` parameter here?

BALSAMIC/commands/run/analysis.py
```diff
@@ -93,9 +93,12 @@
...
+    f'Run workflow with selected variant caller(s) disable. Use comma to remove multiple variant callers. Valid '
+    f'valules are: {list(VCF_DICT.keys())}',
```
ashwini06, line 100:
```suggestion
    f'values are: {list(VCF_DICT.keys())}',
```

ashwini06 (commented):
just one general comment

BALSAMIC/commands/run/analysis.py
hassanfa: Good point! Default is string but it's good to add it.

hassanfa: Other Click options for Click.str() are not set, for sake of consistency I keep it the same here as well. Hope it's ok.

PR #438, opened 2020-10-09
refactor reference, remove run reference

hassanfa:
### This PR:

Added:
- `balsamic init` to initialize balsamic
- `balsmaic init reference` to create and download reference
- Ability to download and update references together

Changed:
- Reference directory is now reflects balsamic version: e.g. `reference/6.0.0/hg38`

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


hassanfa:
`--quiet` works and reference download works:
image

Output directories now structured per reference:
image

hassanfa:
Ability to download and update references independently:
image

BALSAMIC/utils/cli.py
```diff
@@ -56,6 +56,7 @@ class SnakeMake:
     run_analysis    - To run pipeline
     use_singularity - To use singularity
     singularity_bind- Singularity bind path
+    quiet           - Queit mode for snakemake
```
ashwini06, line 59:
```suggestion
    quiet           - Quiet mode for snakemake
```

BALSAMIC/commands/init/reference.py
```diff
@@ -109,3 +134,29 @@ def reference(context, outdir, cosmic_key, snakefile, dagfile, singularity,
...
+    # Construct snakemake command to run workflow
+    balsamic_run = SnakeMake()
+    balsamic_run.working_dir = config['output']
+    balsamic_run.snakefile = snakefile
+    balsamic_run.configfile = config_json
+    balsamic_run.run_mode = "local"
+    balsamic_run.forceall = force_all
+    balsamic_run.run_analysis = run_analysis
+    balsamic_run.quiet = quiet
+    balsamic_run.sm_opt = list(snakemake_opt) + ["--cores", "1"]
```
ashwini06, line 156: how about the `--dag` option? do we need to explicitly mention here or is it left to our choice?

BALSAMIC/commands/init/reference.py
```diff
@@ -46,9 +45,33 @@
...
               help=("Genome version to prepare reference. Path to genome"
                     "will be <outdir>/genome_version"))
+@click.option(
+    '-r',
+    '--run-analysis',
+    show_default=True,
+    default=False,
+    is_flag=True,
+    help=("By default balsamic run_analysis will run in dry run mode."
+          "Raise thise flag to make the actual analysis"))
```
ashwini06, line 55:
```suggestion
          "Raise this flag to make the actual analysis"))
```

ashwini06 (commented):
Looks good👍  just a few typo comments

BALSAMIC/commands/init/reference.py
hassanfa: dag is handled some lines above that I haven't changed in this PR:

```Python
    with CaptureStdout() as graph_dot:
        snakemake.snakemake(snakefile=snakefile,
                            dryrun=True,
                            configfiles=[config_json],
                            printrulegraph=True)
```

Or do you mean we should remove that and add `--dag` here?

We do a lot of dag generation throughout balsamic via snakemake. Which I hope to refactor in the future, and move all of them to a `subprocess` call instead of using `CaptureStdout` class.

hassanfa:
Merging!

PR #439, opened 2020-10-12
Feat/disable var caller report

hassanfa:
### This PR:

Added:
- Disable variant caller for report command. This is to harmonize between workflow and delivery files.

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


hassanfa:
@ashwini06 added bunch of dummy files for status check test. `helpers` fixture needs re-work though. All these dummy files can be handled via fixtures. I'll refactor those when times are less busy :-)

This is to address production's need for disabling variant caller to prepare housekeeper delivery file.

ashwini06 (commented):
👍

ashwini06:
> @ashwini06 added bunch of dummy files for status check test. `helpers` fixture needs re-work though. All these dummy files can be handled via fixtures. I'll refactor those when times are less busy :-)
> 
> This is to address production's need for disabling variant caller to prepare housekeeper delivery file.

@hassanfa Let me know if you need any help in this refactoring

PR #462, opened 2020-10-19
annotate with gnomad_genome

hassanfa:
### This PR:

Added:
- annotate somatic variants with gnomad

Changed:
- extend fitlers for pop_freq using gnomad maxaf

### Review and tests: 
- [ ] Tests pass
- [ ] Code review

### DIFF
--- a/BALSAMIC/assets/vcfanno/vcfanno.toml
+++ b/BALSAMIC/assets/vcfanno/vcfanno.toml
@@ -1,0 +1,5 @@
+[[annotation]]
+file="gnomad.genomes.r2.1.1.sites.vcf.bgz"
+fields = ["AF", "AF_popmax"]
+ops=["self", "self"]
+names=["GNOMADAF", "GNOMADAF_popmax"]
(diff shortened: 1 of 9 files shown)


BALSAMIC/snakemake_rules/annotation/vep.rule
```diff
@@ -4,6 +4,7 @@
 
 from BALSAMIC.utils.rule import get_conda_env
 from BALSAMIC.utils.rule import get_threads
+from BALSAMIC.utils.constants import VCFANNO_TOML
```
ashwini06, line 8:
```suggestion
  singularity: singularity_image
```
How about defining one common singularity image for all rules, instead of mentioning it for every rule inside `vep.rule`

BALSAMIC/assets/vcfanno/vcfanno.toml
```diff
@@ -0,0 +1,5 @@
+[[annotation]]
```
ashwini06, line 1: Any specific reasons to have this as '.toml' file, unlike the rest of YAML files in BALSAMIC.

ashwini06 (commented):
Nice👍

BALSAMIC/snakemake_rules/annotation/vep.rule
hassanfa: I thought when we decouple images, then we can use one image per rule.

BALSAMIC/assets/vcfanno/vcfanno.toml
hassanfa: It is a file that vcfanno needs: https://github.com/brentp/vcfanno#usage

PR #469, opened 2020-10-21
Feat/filterconsensus called reads

ashwini06:
### This PR is a combination of changes:

Added:

- [x] Bash script to filter Consensuscalled reads based on XZ tag. By default, the parameter `min read` is to set to `'3,1,1'` for filtering. 
       - Helps to reduce the number of false-positives from variant-calling. It still needs to identify the best `minreads` settings and check for recall/precision values.

- [x] Refactoring of `sentieon_consensuscall` rule in UMI_workflow
         - Restructured snakemake rule structure and defining the names. Updated cluster.json for the correct rule name
         - Removed the declaration of parameter values in rules. Instead, moved declaration to constant.py,  

## Screenshots
f1
f2

## Review and tests: 
- [ ] Tests pass
- [ ] Code review


hassanfa:
Let me know when it is ready to review.

ashwini06:
> Let me know when it is ready to review.

@hassanfa Now it's ready for review

BALSAMIC/assets/scripts/FilterDuplexUMIconsensus.sh
```diff
@@ -0,0 +1,51 @@
...
+output_bam=$2
+min_reads=$3
+
+# min_reads has three values, e.g. 2,1,1
+# The first value is applies to the total group size,
+# the second value to one single-strand consensus, 
+# and the last value to the other single-strand consensus.
+
+samtools view -h $input_bam | \
+awk -v MinR=$min_reads -v OFS="\t" '
```
hassanfa, line 21: Could we add this as awk script? We can then move rest of it to snakemake shell. 

Reason is, this script has to be moved into a python, R, or Go script for maintainability in future.

BALSAMIC/config/cluster.json
```diff
@@ -186,5 +186,21 @@
 	"vardict_tumor_only": {
 		"time": "10:00:00",
 		"n": 10
+	},
+	"sentieon_umi_align": {
```
hassanfa, line 190: Do UMI rules need 24:00:00? Let's reduce this to a lower number. What do you recommend?

BALSAMIC/snakemake_rules/umi/generate_AF_tables.rule
```diff
@@ -1,27 +1,36 @@
...
 # Generate tables for AF scatterplots
 rule calculate_AF:
     input:
-        vcf = vcf_dir + '{sample}.{var_caller}.umi.vcf.gz'
+        vcf = vcf_dir + '{sample}.{var_caller}.consensusfiltered.vcf.gz'
     output:
-        AF = table_dir + '{sample}.{var_caller}.umi.AFtable.txt'
+        AF = table_dir + '{sample}.{var_caller}.consensusfiltered.AFtable.txt'
+    benchmark:
+        benchmark_dir + '{sample}.{var_caller}.AFcalculate.tsv'
```
hassanfa, line 16: Could we have rule name first for benchmark file.

BALSAMIC/snakemake_rules/umi/sentieon_consensuscall.rule
```diff
@@ -2,55 +2,98 @@
 # coding: utf-8
 
 
+from BALSAMIC.utils.rule import get_conda_env
+from BALSAMIC.utils.rule import get_script_path
 from BALSAMIC.utils.rule import get_threads
+from BALSAMIC.utils.models import UMIworkflowConfig
```
hassanfa, line 8: Wonderful!

BALSAMIC/utils/models.py
```diff
@@ -480,3 +480,18 @@ def validate_path(cls, value, values, **kwargs):
                 output_value = value
 
         return output_value
+
+
+class UMIworkflowParams(BaseModel):
```
hassanfa, line 485: Please add attributes to docstring as well.

BALSAMIC/utils/models.py
```diff
@@ -480,3 +480,18 @@ def validate_path(cls, value, values, **kwargs):
...
+    """This class defines the params settings used as constants in UMI workflow """
+    align_format: str
+    filter_minreads: str = '3,1,1'
+    tag: str = 'XR'
+    align_header: str
+    align_intbases: int
+
+class UMIworkflowConfig(BaseModel):
```
hassanfa, line 493: Please extend docstring.

hassanfa (changes requested):
✨✨
Some minor comments.

hassanfa:
Forgot to mention, please also add tests for two new models you've added in https://github.com/Clinical-Genomics/BALSAMIC/blob/master/tests/utils/test_models.py

BALSAMIC/assets/scripts/FilterDuplexUMIconsensus.sh
ashwini06: yes changed to awk script.

BALSAMIC/snakemake_rules/umi/generate_AF_tables.rule
ashwini06: Fixed

hassanfa (approved):
Very nice! ⭐  ⭐  I liked addition of the parameters to constants. At some point we need to move all rule specific params to constants.

ashwini06:
> Very nice! ⭐ ⭐ I liked addition of the parameters to constants. At some point we need to move all rule specific params to constants.

Thank you :) Sure, will fix in other PR

PR #488, opened 2020-11-05
DRAGEN addition

hassanfa:
### This PR:

Added:
- `--dragen` option to `balsamic run analysis` for tumor only WGS runs

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


hassanfa:
## Running tumor only WGS:
DRAGEN as part of the tumor only WGS workflow:
image

Without `--dragen` it won't be part of the WGS workflow:
image

## DRAGEN will not be running tumor normal WGS:
DRAGEN won't be running as part of TN_WGS workflow:
image

With `--dragen` and a non-WGS case DRAGEN won't run:
## DRAGEN will not be running tumor only panel:
image

## DRAGEN will not be running tumor-normal panel:
image

ashwini06 (approved):
Looks good 👍  🥇  :star:

PR #502, opened 2020-11-11
Feat/add tests umiworkflow

ashwini06:
### This PR:

Tests to check background variant file and run the analysis for UMI workflow


hassanfa (approved):
Aha! That's smart to feed keys to snakemake via a secondary config file! 🌟 

Gave me the idea to do the same within BALSAMIC!

PR #511, opened 2020-11-19
Decouple containers into their own definition

hassanfa:
### This PR:

Changed:
- Decouple and move away from single container to multiple container

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


ashwini06 (commented):
Nice 👍
 Can you also include sentieon installation to docker?

hassanfa:
> Nice 👍
> Can you also include sentieon installation to docker?

Good point! I'll add that as well. I think we can add all three versions: 2018, 2019, and 2020. What do you think?

ashwini06:
> > Nice 👍
> > Can you also include sentieon installation to docker?
> 
> Good point! I'll add that as well. I think we can add all three versions: 2018, 2019, and 2020. What do you think?

Cool. But are we using the 2018 version for any specific analysis?

hassanfa:
> > > Nice 👍
> > > Can you also include sentieon installation to docker?
> > 
> > Good point! I'll add that as well. I think we can add all three versions: 2018, 2019, and 2020. What do you think?
> 
> Cool. But are we using the 2018 version for any specific analysis? Maybe just 2019 and 2020 versions isn't good enough?

Riiiight. Running correct version might be tricky then. Also it needs to connect to license server and requires an open port, iirc.

hassanfa:
Mergin' to test develop build/push

PR #514, opened 2020-11-25
Refactor/umi wildcards

ashwini06:
### This PR:
Due to differences in the wildcards between the main balsamic workflow and UMI workflow, it is not possible to use some of the rules in a mix-match way.  
- In this PR, the existing wildcard `{sample}` is changed to `{casename}` which helps to utilize annotate rule from main balsamic rules

**Added:**
- ` vcfanno` will be run on the umi generated vcf files from now on.

**Changed:** 
- wildcard `{sample}` to `{case_name}`

**Removed:** 
-  `annotate_vep.rule`. Main balsamic `snakemake_rules/annotation/vep.rule` is used instead for annotation
- All redundant variables in individual rules, which are moved to main UMI workflow

**ScreenShot:** Usage of `vep_somatic` rule in UMI workflow run

PR_image

### Review and tests: 
- [x] Tests pass
- [x] Code review


ashwini06:
At the moment, we have wildcard `{step}` that considers two different vcf files (with and without minread filtering). And vep_somatic gets 4 input vcf files (2 from TNscope, 2 from vardict), and outputs 4 annotated vcf files. In the future, if the filtering strategy looks optimal, we can enforce the UMI workflow to consider only `*.consensusfiltered.vcf` file and can skip `{step}`

hassanfa (approved):
👍🏽

PR #535, opened 2021-01-07
Hotfix, reduce MQ filter cutoff to 40

hassanfa:
### This PR:

Fixed:
- Reduces MQ cutoff from 50 to 40 to only remove obvious artifacts.
- Reduces AF cutoff from 0.02 to 0.01

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


ashwini06 (approved):
👍

PR #562, opened 2021-02-17
Feat/umi tn

ashwini06:
### This PR:

For panel datasets, and for both tumor-only and tumor-normal analysis, UMI workflow by default is run along with the balsamic runs.

Fixed:
- wildcards from {case} to {sample}

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


ashwini06:
wrong push to master branch

PR #563, opened 2021-02-17
Feat/umi tn

ashwini06:
### This PR 
Includes refactoring of UMI related snakemake rules and integration to main Balsamic
For panel datasets for both tumor-only and tumor-normal analysis, UMI workflow by default is run along with the balsamic runs.

**Changes:**
- use of  wildcards from {case} to {sample}
- use of only consensusfiltered bam file for variant calling in TNscope

**Removed:**
- Removed umi related rules `bcftools_query_calculatenoiseAF_umi` and `seaborn_densityplot_umi` which are used for calculating noiseAF with consensuscalled and consensfiltered vcf files, and the same for plotting density graphs.
- Removed script function related to `get_densityplot` and its test-related data files and codes from tests/

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


hassanfa (approved):
💯  Commence merge!

PR #597, opened 2021-03-22
Fix deliveries

hassanfa:
### This PR:

Changed:
- rankscore is now a research tag instead of clinical
- Some typo and fixes in the coverage and constant metrics
- Delivery process is more verbose

Fixed:
- CNVKit output is now properly imported in the deliveries and workflow

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


BALSAMIC/utils/rule.py
```diff
@@ -175,11 +180,13 @@ def get_rule_output(rules, rule_name, output_file_wildcards):
                 tags=base_tags,
                 output_file_wildcards=output_file_wildcards)
 
+
             # Return empty string if delivery_id is not resolved.
             # This can happen when wildcard from one rule tries to match with a file
             # from another rule. example: vep_somatic might pick up ngs_filter_vardict files
             pattern = re.compile(r"{([^}\.[!:]+)")
             if pattern.findall(delivery_id):
+                LOG.error("Problem in pattern matching the following: {}".format(delivery_id))
```
ashwini06, line 189: Can you fix tests for the latest additions of log messages?

ashwini06 (commented):
Looks good 👍 Codecoverage dropped a lot, if you can fix that, I think its ready for merge

BALSAMIC/utils/rule.py
hassanfa: I fixed code coverage for recent log messages. The `delivery_id` tags need a bit more complex tests. Overall coverage is increased but I'll work in this in another PR if you don't mind.

ashwini06 (approved):

PR #598, opened 2021-03-24
Refactor/small umi fixes

ashwini06:
### This PR:

- Target coverage for UMI files is wrongly calculated in Picard HSmetrics. Set Coverage_cap to 500.
- Previously, AFtables are calculated using formulae (AF=allelic depth of Alt. allele / (Allelic depths for the Ref + Alt. alleles)). Now, this is modified and directly TNscope calculated AF values are fetched and written to AFtable.txt.


hassanfa (approved):
👍🏽

PR #616, opened 2021-05-05
add tumor lod option for passing final variants

ashwini06:
### This PR:

Variant calling with TNScope in UMIworkflow, tumor lod option uses the default (value 6.2) for passing the final called variants. Now added this parameter to the command-line, and the default value can be controlled via constants.

Tumor LOD default values for tumor and normal in SENTIEON are set according to the mutect. Read more [here](https://software.broadinstitute.org/cancer/cga/mutect)


hassanfa (approved):
💯

PR #635, opened 2021-05-20
Include all filtered variants in vcf-filtered-clinical not just PASSed

hassanfa:
### This PR:

Fixed:

- vcf-filtered-clinical tag only had PASSed variants with further filters. Fixed such it also preserves VarDict filters and not just `PASS`

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


ashwini06 (approved):
👍

PR #648, opened 2021-06-07
Feat/add Delly container

khurrammaqbool:
### This PR:

Added: Delly container

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


hassanfa (approved):
As mentioned during meeting, the base is from `add_SVtools` and not develop. So this PR is into `add_SVtools` instead of `develop`.

Issue #661, opened 2021-06-30
[ Failed Analysis ] vep_annotation out of memory error


hassanfa:
**Case or sample information**
Internal sample id or case id: soughtskinke
Amount read-pairs: 221Mil
Sequence type (WES, TGA, WGS): WGS-tumor-only

**Error message**
```
slurmstepd: error: Job 1366002 exceeded memory limit (44157190144 > 43805310976), being killed
slurmstepd: error: Exceeded job memory limit
slurmstepd: error: *** JOB 1366002 ON compute-0-2 CANCELLED AT 2021-06-29T14:46:53 ***
```

**Link to the error message**
`/home/proj/production/cancer/cases/soughtskink/logs.1/BALSAMIC.soughtskink.vep_somatic.5.sh_1366002.err`

**Reason for failure**
Out of memory due to trying to annotate SVs with gnomad

**If workflow, which rules**
vep_somatic

**Version (please complete the following information):**
7.2.2

**Additional context**
None.

PR #662, opened 2021-06-30
increase cpu resources for vep_somatic

hassanfa:
### This PR:

Fixed:

- Increased CPU resources for `vep_somatic` rule

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


hassanfa:
A proper fix for issue #661 is to separate VEP_somatic for SV and SNV. But that would require changing deliveries. It'll be a minor release. This, however, will solve the issue and only cause a patch fix.

ashwini06 (approved):
👍

khurrammaqbool (approved):

Issue #51, opened 2018-06-05
Tumor mutational burden


hassanfa:
1. TMB was defined as the number of somatic, coding, base substitution, and indel mutations per megabase of genome examined. https://genomemedicine.biomedcentral.com/articles/10.1186/s13073-017-0424-2
1.1. Non-coding alterations were not counted.
1.2 Alterations listed as known somatic alterations in COSMIC and truncations in tumor suppressor genes were not counted
1.3 Alterations predicted to be germline by the somatic-germline-zygosity algorithm were not counted
1.4 Alterations that were recurrently predicted to be germline in our cohort of clinical specimens were not counted.
1.5 Known germline alterations in dbSNP were not counted.
1.6 To calculate the TMB per megabase, the total number of mutations counted is divided by the size of the coding region of the targeted territory.
1.7 select the table 1 from paper above as a ref for comparison.

2. Tumor mutation burden (TMB), fraction of copy number–altered genome, and gene alterations were compared among patients with DCB and no durable benefit (NDB). http://ascopubs.org/doi/full/10.1200/JCO.2017.75.3384
2.1 in addition to above, copy number alterations were also counted.

PR #677, opened 2021-07-12
Optimize TMB calculation

hassanfa:
### This PR:

Updates:

- Improves TMB calculation for panelseq data

### Review and tests: 
- [ ] Tests pass
- [ ] Code review

### DIFF
--- a/BALSAMIC/snakemake_rules/annotation/vep.rule
+++ b/BALSAMIC/snakemake_rules/annotation/vep.rule
@@ -64,7 +64,7 @@ tabix -p vcf -f {output.vcf_all};
 rm $tmpvcf;
     """
 
-rule vep_stat:
+rule tmb_calculation:
   input:
     vep = vep_dir + "SNV.somatic.{case_name}.{var_caller}.all.vcf.gz",
   output:
@@ -75,7 +75,7 @@ rule vep_stat:
     Path(singularity_image, config["bioinfo_tools"].get("ensembl-vep") + ".sif").as_posix() 
   params:
     conda = config["bioinfo_tools"].get("ensembl-vep"),
-    bcftools_filter_string = "INFO/DP>=50 && FORMAT/AD>=5 && FORMAT/AF>=0.1",
+    af_cutoff = "0.05",
     bed = config["panel"]["capture_kit"] if "panel" in config else "",
     filter_vep_string = "'not COSMIC and not Existing_variation and synonymous_variant'", 
     housekeeper_id = {"id": "{case_name}", "tags": "stat-somatic"},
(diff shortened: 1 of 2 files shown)


BALSAMIC/snakemake_rules/annotation/vep.rule
```diff
@@ -90,11 +90,14 @@ source activate {params.conda};
 
 if [ \"{params.bed}\" == \"\" ]; then region_size=3101.78817; else region_size=$(awk '{{s+=$3-$2}}END{{print s/1e6}}' {params.bed}); fi;
```
ashwini06, line 91:
```suggestion
if [ \"{params.bed}\" == \"\" ]; then region_size=3101.78817; 
```
For panels we always have the bed file right ? Then what is the purpose of pre-assigning this value `3101.78817` when bed file is none ?

ashwini06 (commented):
👍

khurrammaqbool (approved):

hassanfa:
@khurrammaqbool @ashwini06 

TMB calculation is following the method described in the linked issue: https://github.com/Clinical-Genomics/BALSAMIC/issues/51

It is optimized according to this publication as well: https://pubmed.ncbi.nlm.nih.gov/30238975/

For that weird number, have a look at for total genome length and why it is a float number: https://www.ncbi.nlm.nih.gov/grc/human/data?asm=GRCh37.p13

Also note that TMB is not validated yet, but this is an effort to optimize it further until we do some sort of validation.

BALSAMIC/snakemake_rules/annotation/vep.rule
```diff
@@ -75,7 +75,7 @@ rule vep_stat:
     Path(singularity_image, config["bioinfo_tools"].get("ensembl-vep") + ".sif").as_posix() 
   params:
     conda = config["bioinfo_tools"].get("ensembl-vep"),
-    bcftools_filter_string = "INFO/DP>=50 && FORMAT/AD>=5 && FORMAT/AF>=0.1",
```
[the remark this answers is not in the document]
hassanfa: See the comment. There are two publications on how to calculate tmb. TMB is still away from being complete, as AF cutoff should be applied anyway. See the publications and issue in my latest comment.

BALSAMIC/snakemake_rules/annotation/vep.rule
khurrammaqbool: Correction:  Genome size/100000

hassanfa: Divided by 1e6. Total length is 3 101 788 170, thus 3101.78817. Again this is a very crude estimation. The actual size should be calculated from the reference we are using. At least until we validate TMB.

PR #689, opened 2021-07-19
Feat/ascat download container

khurrammaqbool:
### This PR:

Added: ascat download container

### Review and tests: 
- [ ] Tests pass
- [ ] Code review

### DIFF
--- a/BALSAMIC/config/balsamic_env.yaml
+++ b/BALSAMIC/config/balsamic_env.yaml
@@ -29,3 +29,5 @@ vcf_merge:
   - vcfmerge
 varcall_delly:
   - delly
+ascatngs:
+  -ascat
--- a/BALSAMIC/config/balsamic_env.yaml
+++ b/BALSAMIC/config/balsamic_env.yaml
@@ -29,5 +29,3 @@ vcf_merge:
   - vcfmerge
 varcall_delly:
   - delly
-ascatngs:
-  -ascat
--- a/BALSAMIC/utils/constants.py
+++ b/BALSAMIC/utils/constants.py
@@ -595,7 +595,7 @@ umiworkflow_params = {
 # list of bioinfo tools for each conda env
 VALID_CONTAINER_CONDA_NAME = {
     "align_qc", "annotate", "coverage_qc", "varcall_py36", "varcall_py27",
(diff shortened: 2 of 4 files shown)


BALSAMIC/config/balsamic_env.yaml
```diff
@@ -28,4 +28,4 @@ varcall_cnvkit:
 vcf_merge:
   - vcfmerge
 varcall_delly:
-  - delly
+  - delly
```
hassanfa, line 31: Add ascatngs here as well.
```suggestion
  - delly
  ascatngs:
     - ascat
```

BALSAMIC/utils/constants.py
```diff
@@ -619,7 +619,8 @@
     "strelka": "varcall_py27",
     "manta": "varcall_py27",
     "cnvkit": "varcall_cnvkit",
-    "delly": "varcall_delly"
+    "delly": "varcall_delly",
+    "ascat": "ascatngs"
 }
```
hassanfa, line 624: As we discussed `BIOINFO_TOOL_SUBMODULE` should be created here.

```suggestion
}

BIOINFO_TOOL_SUBMODULE = {
    "ascat": "ascatngs"
}
```

And in https://github.com/Clinical-Genomics/BALSAMIC/blob/5b754baf1b2d151dc1fcd25d35b117d36f4bdbeb/BALSAMIC/commands/config/case.py#L12-L13 add `BIOINFO_TOOL_SUBMODULE` here as well.

And merge two dictionaries here:

https://github.com/Clinical-Genomics/BALSAMIC/blob/5b754baf1b2d151dc1fcd25d35b117d36f4bdbeb/BALSAMIC/commands/config/case.py#L134 merge two dictionaries `BIOINFO_TOOL_SUBMODULE` and `BIOINFO_TOOL_ENV`.

Ultimately in the new rule, it'll pickup correct ascatngs container.

hassanfa (changes requested):
As we discussed here are my recommendations.

hassanfa (approved):
Looks good! 💯 complex problem but an easy solution.

hassanfa:
You can merge after the tests are done.

PR #698, opened 2021-07-29
Fix/delly run

khurrammaqbool:
### This PR:

Fixed: Delly run.

### Review and tests: 
- [x] Tests pass
- [x] Code review

### DIFF
--- a/BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_normal.rule
+++ b/BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_normal.rule
@@ -81,13 +81,13 @@ rule delly_tumor_normal:
         """
 source activate {params.conda};
 
-delly call -x {input.excl} -o {params.tmpdir}/delly.bcf -g {input.fa} {input.bamT} {input.bamN}
+delly call -x {input.excl} -o {params.tmpdir}/delly.bcf -g {input.fa} {input.bamT} {input.bamN};
 
-echo TUMOR$'\t'tumor$'\n'NORMAL$'\t'control > {params.tmpdir}/samples.tsv
+echo -e \"{params.tumor}\\ttumor\\n{params.normal}\\tcontrol\" > {params.tmpdir}/samples.tsv;
 
-delly filter -p -f somatic -o {params.tmpdir}/delly.somatic.bcf -s {params.tmpdir}/samples.tsv {params.tmpdir}/delly.bcf
+delly filter -p -f somatic -o {params.tmpdir}/delly.somatic.bcf -s {params.tmpdir}/samples.tsv {params.tmpdir}/delly.bcf;
 
-bcftools view -O z -o {output.final} {params.tmpdir}/delly.somatic.bcf
+bcftools view -O z -o {output.final} {params.tmpdir}/delly.somatic.bcf;
 
 tabix -p vcf -f {output.final};
 
--- a/BALSAMIC/utils/models.py
+++ b/BALSAMIC/utils/models.py
@@ -162,7 +162,7 @@ class VCFModel(BaseModel):
     manta_germline: VarcallerAttribute
     haplotypecaller: VarcallerAttribute
     TNscope_umi: VarcallerAttribute
-    delly: VarcallerAttribute
+#    delly: VarcallerAttribute
 
 class AnalysisModel(BaseModel):
     """Pydantic model containing workflow variables
--- a/BALSAMIC/utils/models.py
+++ b/BALSAMIC/utils/models.py
@@ -162,7 +162,7 @@ class VCFModel(BaseModel):
     manta_germline: VarcallerAttribute
     haplotypecaller: VarcallerAttribute
     TNscope_umi: VarcallerAttribute
-#    delly: VarcallerAttribute
+    delly: VarcallerAttribute
 
 class AnalysisModel(BaseModel):
     """Pydantic model containing workflow variables
--- a/BALSAMIC/utils/models.py
+++ b/BALSAMIC/utils/models.py
@@ -162,7 +162,7 @@ class VCFModel(BaseModel):
     manta_germline: VarcallerAttribute
     haplotypecaller: VarcallerAttribute
     TNscope_umi: VarcallerAttribute
-
+    delly: VarcallerAttribute
 
 class AnalysisModel(BaseModel):
     """Pydantic model containing workflow variables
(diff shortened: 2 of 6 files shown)


BALSAMIC/workflows/balsamic.smk
```diff
@@ -367,6 +367,8 @@ if config["analysis"]["sequencing_type"] == "wgs" and config['analysis']['analys
     if "dragen" in config:
         analysis_specific_results.extend([Path(result_dir, "dragen", "SNV.somatic." + config["analysis"]["case_id"] + ".dragen_tumor.bam").as_posix(),
                                           Path(result_dir, "dragen", "SNV.somatic." + config["analysis"]["case_id"] + ".dragen.vcf.gz").as_posix()])
+# delly get vcf file
```
hassanfa, line 370: This is not needed if you have added it to the models.

In order to have another file produced then you should simply add delly here

https://github.com/Clinical-Genomics/BALSAMIC/blob/9ed348a37893b375ff367aa5498eb212a37424ef/BALSAMIC/workflows/balsamic.smk#L362

hassanfa (changes requested):
Looks like you already managed to fix it, but you don't need to add it separately. Either make it so that vep rule can pick it up or add it to the line i commented.

All the snv and sv somatic is annotated here: https://github.com/Clinical-Genomics/BALSAMIC/blob/9ed348a37893b375ff367aa5498eb212a37424ef/BALSAMIC/workflows/balsamic.smk#L334 irregardless of their source as long as they are in the model.

BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_normal.rule
```diff
@@ -81,13 +81,13 @@ rule delly_tumor_normal:
         """
 source activate {params.conda};
 
-delly call -x {input.excl} -o {params.tmpdir}/delly.bcf -g {input.fa} {input.bamT} {input.bamN}
+delly call -x {input.excl} -o {params.tmpdir}/delly.bcf -g {input.fa} {input.bamT} {input.bamN};
 
-echo TUMOR$'\t'tumor$'\n'NORMAL$'\t'control > {params.tmpdir}/samples.tsv
+echo -e \"{params.tumor}\\ttumor\\n{params.normal}\\tcontrol\" > {params.tmpdir}/samples.tsv;
```
ashwini06, line 86:
```suggestion
echo -e \"{params.tumor}\\tTUMOR\\n{params.normal}\\tNORMAL\" > {params.tmpdir}/samples.tsv;
```

BALSAMIC/utils/models.py
```diff
@@ -162,7 +162,7 @@ class VCFModel(BaseModel):
     manta_germline: VarcallerAttribute
     haplotypecaller: VarcallerAttribute
     TNscope_umi: VarcallerAttribute
-
+#    delly: VarcallerAttribute
```
ashwini06, line 165:
```suggestion
    delly: VarcallerAttribute
```

BALSAMIC/workflows/balsamic.smk
```diff
@@ -367,6 +367,8 @@ if config["analysis"]["sequencing_type"] == "wgs" and config['analysis']['analys
...
+analysis_specific_results.extend([Path(vcf_dir + "SV.somatic." + config["analysis"]["case_id"] + ".delly.vcf.gz")])
```
ashwini06, line 371: Can you add the `vep` annotated `delly` vcf as the final analysis output? And you can combine this with line 363. As we discussed, try using `get_vcf` for obtaining the delly annotated vcf file.

ashwini06 (commented):
👍  Few comments and after fixing these changes, can you also paste a screenshot of balsamic dry-run mode to see the effective changes

BALSAMIC/workflows/balsamic.smk
khurrammaqbool: I have tried, but get_vcf on delly fails.

BALSAMIC/utils/models.py
khurrammaqbool: I have commented it because it failed py.test and I could not figure out from the log. Any suggestion on it.

BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_normal.rule
khurrammaqbool: sample.tsv is not same as namemap, so we generate both files.

BALSAMIC/workflows/balsamic.smk
```diff
@@ -363,12 +363,13 @@ else:
 analysis_specific_results.extend([expand(vep_dir + "{vcf}.all.filtered.pass.vcf.gz",
                                         vcf=get_vcf(config, ["manta"], [config["analysis"]["case_id"]]))])
 
+analysis_specific_results.extend([expand(vcf_dir + "{vcf}.vcf.gz",
```
ashwini06, line 366: You are close enough, but we need annotated delly vcf as file output. I suggest the following

```
analysis_specific_results.extend([expand(vep_dir + "{vcf}.all.filtered.pass.vcf.gz",
                                        vcf=get_vcf(config, ["manta", "delly"], [config["analysis"]["case_id"]]))])
```

But this won't work unless you fix your delly output vcf filename. Take a close on the pattern of your delly rule vcf output and vep input filename and change accordingly. https://github.com/Clinical-Genomics/BALSAMIC/blob/81668fa5aae9fe81c63c637d0b2282546074daa6/BALSAMIC/snakemake_rules/annotation/vep.rule#L7

ashwini06 (commented):
Please check my above comment

khurrammaqbool:
Screenshot 2021-07-30 at 10 53 01
Screenshot 2021-07-30 at 10 54 51

BALSAMIC/workflows/balsamic.smk
khurrammaqbool: Thanks! It is fixed now.

BALSAMIC/workflows/balsamic.smk
khurrammaqbool: It is now fixed.

ashwini06 (approved):
It's good to go 👍  
**Optional** : A small suggestion could be in snakemake delly rules, if params are described in long format rather than short-form, (`eg: -f/--filter`), the quality of code will be enhanced for readability.

Issue #711, opened 2021-08-19
[Refactor]: Move params to constants


ashwini06:
**Is your feature request related to a problem? Please describe.**
In the somatic variant calling rules rules, the same parameters are declared more than once. For eg: 

https://github.com/Clinical-Genomics/BALSAMIC/blob/develop/BALSAMIC/snakemake_rules/variant_calling/somatic_tumor_only.rule#L20-L23

https://github.com/Clinical-Genomics/BALSAMIC/blob/develop/BALSAMIC/snakemake_rules/variant_calling/somatic_tumor_normal.rule#L14-L17

**Describe the solution you'd like**
Move these values to `constants.py` as `balsamicworkflow_params`.

https://github.com/Clinical-Genomics/BALSAMIC/blob/04da86f15db007d1b71056df86f48c5c695a343d/BALSAMIC/utils/constants.py#L569 

**Current BALSAMIC version**
`7.2.6`

PR #714, opened 2021-08-24
Refactor/params to constants

ashwini06:
### This PR:
The same parameters are used across multiple rules. Move all the params to `constants.py` and load them accordingly
Related to #711 

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


hassanfa (approved):
💯 This is a very good effort to use common params, and I don't have a counter argument on the structure. We will use this structure and optimize it overtime. Very nice work!

PR #736, opened 2021-09-01
refactor: rename pon reference filename

ashwini06:
### This PR:

Changed:

Rename the PON reference filename with more descriptive details as from which panel and case_id, the output PON is generated. Related to https://github.com/Clinical-Genomics/cgp-cancer-cnvcall/issues/10

To reproduce:

```
balsamic config pon --case-id securetoad10 -p /home/proj/production/cancer/reference/target_capture_bed/production/balsamic/gmcksolid_4.1_hg19_design.bed --balsamic-cache /home/proj/stage/cancer/balsamic_cache/ --analysis-dir /home/proj/long-term-stage/cancer/PON_analysis_runs_APJ/PON_test_datasets/ --fastq-path /home/proj/production/cancer/cases/securetoad/fastq/ -g hg19

balsamic run analysis -s /home/proj/long-term-stage/cancer/PON_analysis_runs_APJ/PON_test_datasets/securetoad10/securetoad10_PON.json 

```

### Review and tests: 
- [ ] Tests pass
- [ ] Code review

### DIFF
--- a/BALSAMIC/workflows/PON.smk
+++ b/BALSAMIC/workflows/PON.smk
@@ -38,10 +38,21 @@ picarddup = get_picard_mrkdup(config)
 samples = get_pon_samples(fastq_dir)
 
 
+panel_ref_dir = os.path.split(target_bed)[0]
+panel_name = os.path.split(target_bed)[1].replace('.bed','')
+
+pon_dir = panel_ref_dir + "/" + "PON_references" + "/"
+
+if not os.path.exists(pon_directory):
+    os.makedirs(pon_directory)
+
 ALL_COVS = expand(cnv_dir + "{sample}.{cov}coverage.cnn", sample=samples, cov=['target','antitarget'])
 ALL_REFS = expand(cnv_dir + "{cov}.bed", cov=['target','antitarget'])
-ALL_PON = expand(cnv_dir + config["analysis"]["case_id"] + "_PON_reference.cnn")
-PON_DONE = expand(cnv_dir + "PON." + "reference" + ".done")
+ALL_PON = expand(pon_dir + panel_name + "_" + config["analysis"]["case_id"] + "_PON_reference.cnn")
+PON_DONE = expand(pon_dir + panel_name + "_" + config["analysis"]["case_id"] + "_PON_reference.done")
+
+#ALL_PON = expand(cnv_dir + config["analysis"]["case_id"] + "_PON_reference.cnn")
+#PON_DONE = expand(cnv_dir + "PON." + "reference" + ".done")
 
 config["rules"] = ["snakemake_rules/quality_control/fastp.rule", 
                    "snakemake_rules/align/bwa_mem.rule"]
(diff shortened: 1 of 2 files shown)


hassanfa (approved):
👍🏽 

Not that it is important, but I recommend to have constants upper case, and variables lower case. If ALL_* variables are indeed constants, keep them upper otherwise all of them should be lowercase. you make the call.

ashwini06:
@hassanfa : No this doesnt work , no permissions to write to this path `home/proj/production/cancer/reference/target_capture_bed/production/balsamic/PON_references`
Screenshot 2021-09-01 at 15 02 55

BALSAMIC/workflows/PON.smk
```diff
@@ -37,11 +37,17 @@ Path.mkdir(Path(tmp_dir), exist_ok=True)
 picarddup = get_picard_mrkdup(config)
 samples = get_pon_samples(fastq_dir)
 
+panel_ref_dir = os.path.split(target_bed)[0]
+panel_name = os.path.split(target_bed)[1].replace('.bed','')
+pon_dir = panel_ref_dir + "/" + "PON_references" + "/"
```
hassanfa, line 42: I commented on PR. These lines have to be changed. Can we use `--analysis-dir` functionality here? The same way that a result_dir is defined in main workflow, result_dir is used to store results.

hassanfa (changes requested):

BALSAMIC/workflows/PON.smk
ashwini06: @hassanfa: Previously, it was writing to the `result_dir` in `cnv` folder. Please see lines 25, 43-44 in this script.

ashwini06:
> @ashwini06 this should not write in the the same path that we are using for panel of normals. It should write the CNN to a user defined path. And then user can decide what to do with it. Is it possible to change the behavior as such?

But isn't we decided to store all PON references in a commonplace? Related to Clinical-Genomics/cgp-cancer-cnvcall#10

hassanfa:
We discussed it, of course. Probably I didn't describe the solution properly, that is to generate a `*cnn` file then add it to the repository via a pull request and not directly on Hasta.

Does that make sense? This way, we can control what and which capture bed is going to have a panel of normal.

ashwini06:
> We discussed it, of course. Probably I didn't describe the solution properly, that is to generate a `*cnn` file then add it to the repository via a pull request and not directly on Hasta.
> 
> Does that make sense? This way, we can control what and which capture bed is going to have a panel of normal.

@hassanfa : Maybe, earlier I also didn't capture your message correctly :innocent: OK now this is clear. So, the previous implementation in the PON workflow is doing the right thing of storing PON reference `.cnn` file in the results_dir. Now i won't change its storage location, instead in this PR, I will add the panel name to the generated PON reference file. Previously it was just called as `cnv_dir + config["analysis"]["case_id"] + "_PON_reference.cnn"`. New change will be like `cnv_dir + panel_name + "_" + config["analysis"]["case_id"] + "_PON_reference.cnn"`. This will give better idea, from which panel and case_id, the PON_reference.cnn is generated from.

aaa

PR #758, opened 2021-09-14
fix: pip install inside balsamic container

ivadym:
### This PR:

Fixes the pip install misbehaviour of balsamic inside its container.

### Review and tests: 
- [ ] Tests pass
- [ ] Code review

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,11 @@ Added:
 
 * BALSAMIC container #728
 
+Fixed:
+^^^^^^
+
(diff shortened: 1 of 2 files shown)


hassanfa (approved):
Worked for me. quick fix! 💯

ivadym:
Merging the PR.

Updated image in docker hub, and also tested on hasta.
Balsamic version 8.1.0.

**`--upgrade` flag**

`pip --upgrade` (which is equal to `--upgrade-strategy only-if-needed`) updates only the packages that are outdated according to the requirements.

Issue #782, opened 2021-10-01
[Failed Analysis]: TMB calculation failures


ashwini06:
`TNscope UMI` vcf files shouldn't be feed to the TMB rule.

<img width="1387" alt="tmb1" src="https://user-images.githubusercontent.com/6302819/135598855-547959bb-a11f-4a5b-a702-d34205fe6f95.png">
Also errors with TNhaplotyper vcfs
<img width="1378" alt="tmb2" src="https://user-images.githubusercontent.com/6302819/135599106-53ff406a-4914-4a8f-8f7e-a3a6428dfc2f.png">

**Current affected rule if snakemake workflow related**
If possible, describe which rules or workflows might be affected by this.

**References and other implementations**
If possible, an example of similar implementations.

PR #786, opened 2021-10-06
fix: tmb failures

ashwini06:
### This PR:

Fixes: #782 
Changed: Sample names in `cnvkit` output files are changed according to other callers' vcf format.

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


BALSAMIC/snakemake_rules/annotation/vep.rule
```diff
@@ -80,14 +81,34 @@ rule tmb_calculation:
...
+| bgzip -c > {params.tmpfiles}.tumor.txt.gz;
+
+tabix -f -s1 -b2 -e2 {params.tmpfiles}.tumor.txt.gz;
+
+bcftools view -s TUMOR {input.vep} \
+| bcftools annotate -s TUMOR \
+-a {params.tmpfiles}.tumor.txt.gz \
+-h {params.tmpfiles}.header \
+-c CHROM,POS,INFO/AF_TUMOR \
+-O z -o {input.vep}.temp;
```
ashwini06, line 98: @hassanfa: Let me know if this new vcf file needs a proper name. As this intermediate vcf file is not used anywhere I thought after generating tmb output, we can delete it. What do you think?

ashwini06:
@hassanfa : Tested with Tumor only panel case and for variant callers `vardict, tnhaplotyper, tnscope_umi, delly` tmb output file `_balsamic.stat` is successfully created. Whereas for `cnvkit, manta` which has a different vcf format, shows up with this error `Error: no such tag defined in the VCF header: FORMAT/AF`. How about creating an exceptional dummy `_balsamic.stat` file for these 2 callers?

PR #819, opened 2021-11-11
fix: cnvkit gender comparison operator

ivadym:
### This PR:

Fixes the `cnvkit.py` call error made from `cnvkit_single.rule` and `cnvkit_paired.rule` during the validation of `BALSAMIC 8.2.0` on stage using the PANEL_UMI TUMOR_ONLY CASE (`uphippo`).

`uphippo` validation case log:

Screen Shot 2021-11-11 at 14 45 10

### Review and tests: 
- [x] Tests pass
- [x] Code review

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,11 @@
+[8.2.1]
+-------
+
+Fixed:
+^^^^^^
+
+* ``cnvkit`` gender comparison operator bug #819
+
 [8.2.0]
 -------
 
@@ -53,8 +61,6 @@ Fixed:
 * ReadtheDocs building failure due to dependencies, fixed by locking versions #773
 * Dev requirements installation for Sphinx docs (Github Action) #812
 * Changed path for main Dockerfile version in ``.bumpversion.cfg``
-* ``cnvkit`` gender comparison operator bug #819
-
 
 [8.1.0]
 -------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -53,6 +53,7 @@ Fixed:
 * ReadtheDocs building failure due to dependencies, fixed by locking versions #773
 * Dev requirements installation for Sphinx docs (Github Action) #812
 * Changed path for main Dockerfile version in ``.bumpversion.cfg``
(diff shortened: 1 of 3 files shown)


ashwini06 (approved):
Thanks for the fix @ivadym 👍🏼

CHANGELOG.rst
```diff
@@ -53,6 +53,7 @@ Fixed:
 * ReadtheDocs building failure due to dependencies, fixed by locking versions #773
 * Dev requirements installation for Sphinx docs (Github Action) #812
 * Changed path for main Dockerfile version in ``.bumpversion.cfg``
+* ``cnvkit`` gender comparison operator bug #819
```
hassanfa, line 56: Add this part to a new version, since it is a hotfix: 8.2.1

hassanfa (approved):
I suggest to change the PR title to "hotfix: ...." and treat this as a patch release. Then you don't need to create a release branch.

CHANGELOG.rst
ivadym: True, thanks! 💯

ivadym:
> I suggest to change the PR title to "hotfix: ...." and treat this as a patch release. Then you don't need to create a release branch.

It seems that `hotfix: ...` is not a valid PR title ([here](https://github.com/commitizen/conventional-commit-types/blob/master/index.json) are the default types). However, I can modify the GitHub actions to include that type as well.

hassanfa:
> > I suggest to change the PR title to "hotfix: ...." and treat this as a patch release. Then you don't need to create a release branch.
> 
> It seems that `hotfix: ...` is not a valid PR title ([here](https://github.com/commitizen/conventional-commit-types/blob/master/index.json) are the default types). However, I can modify the GitHub actions to include that type as well.

you're absolutely right. Then `fix` it is. 👌🏽

ashwini06:
Failed cnvkit run on uphippo case is tested again with this PR code. CNVkit is successfully finished. 
Related log here `/home/proj/stage/cancer/cases/uphippo/logs.1/BALSAMIC.uphippo.cnvkit_single.72.sh_2023186.err`

@ivadym: You can merge this PR.

PR #825, opened 2021-11-29
fix: cnvkit genderissue

### LINKED ISSUE #824: [Failed]: wrong reporting of gender info from PureCN output
**Describe the bug**
Currently, bioinfo tool `PureCN` is used within cnvkit rules, to get information about purity, ploidy and gender information. 
While running validation case `setamoeba`, cnvkit rule is failed due to the following error.

```
cnvkit.py call: error: argument -x/--sample-sex/-g/--gender: invalid choice: 'coverage:' (choose from 'm', 'y', 'male', 'Male', 'f', 'x', 'female', 'Female')
```
The problem is with the output of `PureCN/TUMOR.csv` file, where the .csv is not properly formatted. 

```
"tumor.initial",0.41,3.37407031443908,"Coverage: M VCF: F",0.0587944328263289,TRUE,FALSE,FALSE,"POTENTIAL SAMPLE CONTAMINATION"
```
WARN : Sex mismatch of coverage and VCF. Could be because of noisy data, contamination, loss of chrY or a mis-alignment of coverage and VCF.

**Solution**
For an immediate fix, let cnvkit calculate the gender info from avg log2 averages of chrX.
More [here](https://cnvkit.readthedocs.io/en/stable/reports.html#sex)

ashwini06:
### This PR:

Fixes #824 

- Possibility to report dual information about gender info in purecn/TUMOR.CSV file. Could be because of noisy data, contamination, loss of chrY, or misalignment of coverage and VCF.
- In such cases, consider gender as default `m`
- Updated `purecn` and its dependencies to latest versions. 

### Review and tests: 
- [ ] Tests pass
- [ ] Code review

### DIFF
--- a/BALSAMIC/containers/varcall_cnvkit/varcall_cnvkit.yaml
+++ b/BALSAMIC/containers/varcall_cnvkit/varcall_cnvkit.yaml
@@ -6,12 +6,12 @@ dependencies:
   - conda-forge::python=3.7
   - conda-forge::pip
   - conda-forge::r-base=4.1.0
-  - conda-forge::r-optparse=1.6.6
-  - bioconda::bioconductor-iranges=2.26.0
-  - bioconda::bioconductor-genomeinfodb=1.28.0
-  - bioconda::bioconductor-genomicranges=1.44.0
-  - bioconda::bioconductor-dnacopy=1.66.0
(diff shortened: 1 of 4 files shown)


ivadym (approved):
👏 👏

PR #829, opened 2021-12-06
Feat/add dog

rannick:
### This PR:

Added: canfam3 references 
Changed: general rule handling for compatibility with canfam3

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


hassanfa (changes requested):
I was not sure if this PR is ready to review. I thought before you spend too much time on it, I help a bit so you are time is not wasted.

Please reduce the size for the following items:
1. FASTQ files: how many reads are they?
    1. Reduce them to 50k reads or something
    2. Since this is dog exome, I suggest to select one chromosome only.
2. Please reduce all the reference files for dog, to size **zero** This is for testing purpose only, and no one will run analysis on it. Check out human reference files for inspiration on which ones should be zero size.
3. Select only part of the bed file, see point 1.2. above and only select part of one of the chromosomes. Checkout human `panel.bed´ as it is only sub-100 regions.


Regarding why I asked to remove all the canfam3 specific additions:
1. Snakemake works on targets and rules.
2. For canfam3, if the goal is only QC: you should create a QC only workflow. This way, you will significantly reduce the amount of work you need to do, and QC only can work for other samples and all the other species as long as they have a reference and FASTQ files.
3. For canfam3, if the goal is only variant call: you should create a stop point or `target` for Snakemake. So instead of creating lots of specific rules and functions and changes, you'll have access to all the Snakemake's powerhouse to control and run what you need.

@rannick Great work! 

PS: I'll convert this to draft, since I think it still is work in progress. Change it back if you disagree.

ashwini06:
This PR will be split into two individual PRs 
a: generating canfam3_reference.json; 
b: balsamic qc_only (QC+align) workflow

PR #840, opened 2021-12-20
fix: execute python script inside balsamic container

ivadym:
### This PR:

Runs the `create_pdf.py` python script from the balsamic container, which should now be downloaded when the balsamic cache is generated.

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


ivadym:
Base is develop, this is a hotfix.

Issue #865, opened 2022-02-21
.cram for UMI consensus called files


ashwini06:
Currently '.bam' files are being generated UMI_consensus_called reads. convert bam to cram for these files

ashwini06:
Fixed in #868 and available in production from `balsamic v8.2.8`

PR #868, opened 2022-02-24
feat: generate umi cram files

ashwini06:
### This PR:

Added: Generates CRAM files from UMI-consensus called bam files. Related to #865 

### Review and tests: 
- [ ] Tests pass
- [ ] Code review

### DIFF
--- a/BALSAMIC/constants/workflow_rules.py
+++ b/BALSAMIC/constants/workflow_rules.py
@@ -136,7 +136,7 @@ DELIVERY_RULES = [
     "mergeBam_normal",
     "mergeBam_tumor_umiconsensus",
     "mergeBam_normal_umiconsensus",
-	"cnvkit_paired",
+    "cnvkit_paired",
     "cnvkit_single",
     "ascat_tumor_normal_merge_output",
 ]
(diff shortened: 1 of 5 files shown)


ashwini06:
@ivadym @khurrammaqbool anyone review this PR? @ivadym: umi cram files should be stored in HK, will you make necessary related changes in Hermes?

khurrammaqbool (approved):

ivadym (changes requested):
Just two things:

-  A housekeeper id should be added to the rules in order for the cram files to be stored. If you add a wildcard to the output file it will be also treated as a tag.
```
params:
        housekeeper_id = {"id": config["analysis"]["case_id"], "tags": " ... "}
```

- If we don't want to store bam files, then the bam output file should be marked as `tmp`, right? I can also just ignore it in Hermes, but with this implementation it will appear in the `.hk` file.

ivadym (approved):
👍

I'll make the corresponding changes in Hermes to store themö

ashwini06:
> 👍
> 
> I'll make the corresponding changes in Hermes to store them.


Thanks @ivadym. For a reason, I didn't mark the `"tumor_umi_consensusfiltered.merged.bam"` and `_consensuscalled.bam` as `temp`. But you know only the UMI cram's should be stored and rest (umi related bams) can be ignored in HK.

PR #872, opened 2022-03-01
feat: add SVDB

khurrammaqbool:
### This PR:

(If this is a release PR, no need to add following. Leave this part empty)
(Use the following lines to create a PR text body. Make sure to remove all non-relevant one after you're done)
(Repeat each field as many times as necessary)

Added: SVDB to varcall_py36 container.

### Review and tests: 
- [ ] Tests pass
- [ ] Code review

### DIFF
--- a/BALSAMIC/config/balsamic_env.yaml
+++ b/BALSAMIC/config/balsamic_env.yaml
@@ -19,6 +19,7 @@ varcall_py36:
   - gatk
   - vardict
   - libiconv
+  - svdb
 varcall_py27:
   - strelka
   - manta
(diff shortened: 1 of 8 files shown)


khurrammaqbool:
> 👍
> 
> And as we discussed, since the python version has been updated to `3.8`, the name of the container should be changed in the future to `varcall_py38` or `varcal_py3`.



> 👍
> 
> And as we discussed, since the python version has been updated to `3.8`, the name of the container should be changed in the future to `varcall_py38` or `varcal_py3`.

I will take it to another PR for both varcall_py36 and varcall_py27

PR #877, opened 2022-03-07
fix: docs importing C dependant libraries

ivadym:
### This PR:

Adresses `ModuleNotFoundError: No module named 'numpy'` when building the docs:
https://readthedocs.org/projects/balsamic/builds/16287901/

Solution from:
https://read-the-docs.readthedocs.io/en/latest/faq.html#i-get-import-errors-on-libraries-that-depend-on-c-modules
https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_mock_imports

### Review and tests: 
- [ ] Tests pass
- [ ] Code review

### DIFF
--- a/.github/workflows/sphinx_docs.yml
+++ b/.github/workflows/sphinx_docs.yml
@@ -18,5 +18,5 @@ jobs:
         uses: ammaraskar/sphinx-action@master
         with:
           docs-folder: "docs/"
-          pre-build-command: "pip install -r requirements-dev.txt"
+          pre-build-command: "pip install -r requirements-dev.txt && python setup.py install"
           build-command: "sphinx-build -T -E -b html -d _build/doctrees-readthedocs -D language=en . _build/html"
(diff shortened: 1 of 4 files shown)


ashwini06 (approved):
👍🏼

ivadym:
I also added `setup.py` requirements installation to the sphinx github action

ashwini06:
> I also added `setup.py` requirements installation to the sphinx github action

Looks like all setup.py packages are properly installed with sphinx git action. But there are also differences in the sphinx versions used by our docs/requiremen.txt (sphinx==4.2.0) and the one used in git actions (https://github.com/ammaraskar/sphinx-action/blob/master/.travis.yml). Does this something to think about?

ivadym:
I don't think so because the problem is actually not with Sphinx, but with compiling the Sphinx documentation with ReadTheDocs, which does not allow having `numpy` as a requirement:
> ...
This happens because our build system doesn’t have the dependencies for building your project. This happens with things like libevent, mysql, and other python packages that depend on C libraries. **We can’t support installing random C binaries on our system**, ...
...

PR #881, opened 2022-03-09
fix: dock build mocking cyvcf dependency

ivadym:
### This PR:

Mocks cyvcf2 import  to build documentation.

Tested successfully in a fork repository ([ivadym/BALSAMIC](https://github.com/ivadym/BALSAMIC)):
https://readthedocs.org/projects/balsamic-docs/builds/16310624/

### Review and tests: 
- [x] Tests pass
- [ ] Code review

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -4,6 +4,11 @@
 # Required
 version: 2
 
+build:
+  os: "ubuntu-20.04"
+  tools:
+    python: "3.7"
+
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
   configuration: docs/conf.py
@@ -11,9 +16,8 @@ sphinx:
 # Optionally build your docs in additional formats such as PDF and ePub
 formats: all
 
-# Optionally set the version of Python and requirements required to build your docs
+# Requirements to build the docs
 python:
-  version: 3.7
   install:
     - requirements: docs/requirements.txt
     - requirements: requirements-dev.txt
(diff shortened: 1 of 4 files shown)


.readthedocs.yml
```diff
@@ -4,16 +4,20 @@
 # Required
 version: 2
 
+build:
+  os: "ubuntu-20.04"
```
ashwini06, line 8: Is this a crucial option for readthedocs successful build?

docs/conf.py
```diff
@@ -48,7 +48,7 @@
 exclude_patterns = ["_build", "Thumbs.db", ".DS_Store"]
 
 # Modules to be mocked up
-autodoc_mock_imports = ["numpy"]
+autodoc_mock_imports = ["cyvcf2"]
```
ashwini06, line 51: so its cyvcf2 not numpy 🐞

ashwini06 (approved):
Good job @ivadym 🥇 💯

.readthedocs.yml
ivadym: Its not, it will use it either way (it's the only one supported and the default one) but I think it was convenient to include it, following the documentation that you shared with me yesterday: https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html#id3

docs/conf.py
ivadym: Seems like it :)

.readthedocs.yml
ashwini06: Alright, good to include it then. As it also gives us more clarity about the default os being used by rtd :)

ivadym: 👍 

I will merge it then!

PR #886, opened 2022-03-14
feat: svdb merge sv cnv

khurrammaqbool:
### This PR:

Added: svdb merge SV and CNV

### Review and tests: 
- [ ] Tests pass
- [ ] Code review

### DIFF
--- a/BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_normal.rule
+++ b/BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_normal.rule
@@ -209,8 +209,12 @@ python {params.merge_ascat_output_script} {output.ascat_output_pdf} {input.sampl
 
 rule svdb_merge_tumor_normal:
     input:
-        manta_vcf = vcf_dir + "SV.somatic." + config["analysis"]["case_id"] + ".manta.vcf.gz",
-        delly_vcf = vcf_dir + "SV.somatic." + config["analysis"]["case_id"] + ".delly.vcf.gz",
+        vcf = expand(
+                vcf_dir + "SV.somatic." + config["analysis"]["case_id"] + ".{caller}.vcf.gz",
+                caller=somatic_caller_sv) +
+              expand(
+                vcf_dir + "CNV.somatic." + config["analysis"]["case_id"] + ".{caller}.vcf.gz",
+                caller=somatic_caller_cnv)
     output:
         svdb_vcf = vcf_dir + "SV.somatic." + config["analysis"]["case_id"] + ".svdb.vcf.gz",
         namemap = vcf_dir + "SV.somatic." + config["analysis"]["case_id"] + ".svdb.sample_name_map",
@@ -222,6 +226,8 @@ rule svdb_merge_tumor_normal:
         tumor = get_sample_type(config["samples"], "tumor"),
         normal = get_sample_type(config["samples"], "normal"),
         case_name = config["analysis"]["case_id"],
+        vcf= lambda wildcards, input:[input[index] + ":" + sv_callers[index] for index in range(0,len(input))],
+        svdb_priority= ",".join(sv_callers)
     threads:
         get_threads(cluster_config, "svdb_merge_tumor_normal")
     message:
@@ -229,9 +235,9 @@ rule svdb_merge_tumor_normal:
     shell:
         """
 svdb --merge --no_intra --bnd_distance 5000 --overlap 0.80 \
---vcf {input.manta_vcf}:manta {input.delly_vcf}:delly \
---priority manta,delly | \
-bgzip -l 9 -c  > {output.svdb_vcf}
+--vcf {params.vcf} \
+--priority {params.svdb_priority} | \
+bgzip -l 9 -c  > {output.svdb_vcf};
 
 echo -e \"{params.tumor}\\tTUMOR\\n{params.normal}\\tNORMAL\" > {output.namemap};
         """
--- a/BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_normal.rule
+++ b/BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_normal.rule
@@ -226,7 +226,7 @@ rule svdb_merge_tumor_normal:
         tumor = get_sample_type(config["samples"], "tumor"),
         normal = get_sample_type(config["samples"], "normal"),
         case_name = config["analysis"]["case_id"],
-        vcf = get_svdb_input(input, svdb_sv_callers_to_merge_prio),
+        vcf= lambda wildcards, input:[input[index] + ":" + svdb_sv_callers_to_merge_prio[index] for index in range(0,len(input))],
         svdb_priority= ",".join(svdb_sv_callers_to_merge_prio)
     threads:
         get_threads(cluster_config, "svdb_merge_tumor_normal")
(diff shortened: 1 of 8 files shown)


BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_normal.rule
```diff
@@ -222,16 +226,18 @@ rule svdb_merge_tumor_normal:
         tumor = get_sample_type(config["samples"], "tumor"),
         normal = get_sample_type(config["samples"], "normal"),
         case_name = config["analysis"]["case_id"],
+        vcf= lambda wildcards, input:[input[index] + ":" + sv_callers[index] for index in range(0,len(input))],
+        svdb_priority= ",".join(sv_callers)
```
hassanfa, line 230: I'd suggest to have this variable named properly. e.g. `svdb_sv_caller_prio` or something.

BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_normal.rule
```diff
@@ -222,16 +226,18 @@ rule svdb_merge_tumor_normal:
         tumor = get_sample_type(config["samples"], "tumor"),
         normal = get_sample_type(config["samples"], "normal"),
         case_name = config["analysis"]["case_id"],
+        vcf= lambda wildcards, input:[input[index] + ":" + sv_callers[index] for index in range(0,len(input))],
```
hassanfa, line 229: Probably create a function for this for readability. A function that takes sv_callers and input vcf, and outputs the string that you want. This way, you'd avoid any potential bug if order of input/sv_callers changes.

BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_only.rule
```diff
@@ -118,16 +122,17 @@ rule svdb_merge_tumor_only:
     params:
         tumor = get_sample_type(config["samples"], "tumor"),
         case_name = config["analysis"]["case_id"],
+        vcf= lambda wildcards, input:[input[index] + ":" + sv_callers[index] for index in range(0,len(input))],
+        svdb_priority= ",".join(sv_callers)
```
hassanfa, line 126: Same comments as above.

BALSAMIC/utils/rule.py
```diff
@@ -75,7 +75,7 @@ def get_variant_callers(
         WorkflowRunError if values are not valid
     """
 
-    valid_variant_callers = set()
+    valid_variant_callers = list()
```
hassanfa, line 78: Changing from list to set might cause redundant entries :-) Make sure you remove them and order it properly now.

BALSAMIC/workflows/balsamic.smk
```diff
@@ -170,6 +172,22 @@ for m in MUTATION_TYPE:
...
+                                            mutation_class="somatic")
+
+somatic_caller_cnv = get_variant_callers(config=config,
+                                            analysis_type=config['analysis']['analysis_type'],
+                                            workflow_solution="BALSAMIC",
+                                            mutation_type="CNV",
+                                            sequencing_type=config["analysis"]["sequencing_type"],
+                                            mutation_class="somatic")
+somatic_caller_sv.remove("svdb")
+sv_callers = somatic_caller_sv + somatic_caller_cnv
```
hassanfa, line 189: As suggested above, rename this properly.

hassanfa (commented):
Brilliant solution. See my suggestions below.

BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_normal.rule
khurrammaqbool: This seems to be the recommended solution as in the [documentation](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html). All other solutions indeed require creating a function, but that bypasses the parts of the rule and create redundancy in the code. The order is dependent on the `valid_variant_callers` from already defined function. A new function will have the same issue unless variant callers are hard coded, which I tried to avoid.

khurrammaqbool: The above comment is related to the second part of the comment. Regarding readability, I will move it from here! :-)

khurrammaqbool: The solution above is currently the only one working without any errors!

BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_normal.rule
```diff
@@ -222,16 +227,19 @@ rule svdb_merge_tumor_normal:
         tumor = get_sample_type(config["samples"], "tumor"),
         normal = get_sample_type(config["samples"], "normal"),
         case_name = config["analysis"]["case_id"],
+        vcf= lambda wildcards, input:[input[index] + ":" + svdb_sv_callers_to_merge_prio[index] for index in range(0,len(input))],
```
ashwini06, line 230: svdb_sv_callers_to_merge_prio - maybe this name can be simplified.

decoupling this into a separate function could simplify the readability. But maybe it will turn the difficulty of passing input vcf into the function without using wildcards. We can leave this as such for now., as it works fine

BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_only.rule
```diff
@@ -118,16 +122,17 @@ rule svdb_merge_tumor_only:
     params:
         tumor = get_sample_type(config["samples"], "tumor"),
         case_name = config["analysis"]["case_id"],
+        vcf= lambda wildcards, input:[input[index] + ":" + svdb_sv_callers_to_merge_prio[index] for index in range(0,len(input))],
```
ashwini06, line 125: same as above

ashwini06 (approved):
Looks good 👍🏼

BALSAMIC/utils/rule.py
hassanfa: This can be ambiguous if/when orders change. Just something to keep it in mind.

BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_normal.rule
```diff
@@ -222,16 +226,18 @@ rule svdb_merge_tumor_normal:
         tumor = get_sample_type(config["samples"], "tumor"),
         normal = get_sample_type(config["samples"], "normal"),
         case_name = config["analysis"]["case_id"],
+        vcf= lambda wildcards, input:[input[index] + ":" + svdb_callers_prio[index] for index in range(0,len(input))],
```
hassanfa, line 229: Shouldn't this work?

```python
def construct_vcf(w, i, sv):
  outp=list()
  for index in range(0,len(i)):
     outp.append(f"i[index]:sv[index]")
  return outp
```

hassanfa (commented):
in the function you added, you forgot to add wildcards as input, no?

otherwise, nicely done.

BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_normal.rule
hassanfa: or maybe not. wildcard is will cause an issue. Ignore this :-)

khurrammaqbool: > or maybe not. wildcard is the limiting factor, and can cause an issue. Ignore this :-)

Also, input is not a list and will not iterate in for....

PR #891, opened 2022-03-17
feat: remove vcfanno and cosmic for sv

khurrammaqbool:
### This PR:

Removed: vcfanno for sv

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


khurrammaqbool:
> If the purpose is to just remove vcfanno for SV varcall results, instead of having another extra rule I suggest having a conditional statement to handle the vcfanno command
I have mentioned separating SV from SNV, while optimizing vep.rule, to be able to allocate different cluster resources for SV and SNV, which is added by splitting the rule along with disabling vcfanno for SV.

> How about CNV annotations?
As discussed in the meetings, we will not annotate every SV and CNV caller output. Only merged SV and CNV will be annotated and delivered.

ashwini06:
How about having a single vep rule instead of splitting them, as I mentioned in the previous comment. Also, can balsamic now can handle the required vep results correctly https://github.com/Clinical-Genomics/BALSAMIC/blob/ebbf99d19b46b479b3e5fbbe2027edcc05c6cda4/BALSAMIC/workflows/balsamic.smk#L240 ??

khurrammaqbool:
> > How about having a single vep rule instead of splitting them, as I mentioned in the previous comment. Also, can balsamic now can handle the required vep results correctly??
> > https://github.com/Clinical-Genomics/BALSAMIC/blob/ebbf99d19b46b479b3e5fbbe2027edcc05c6cda4/BALSAMIC/workflows/balsamic.smk#L240
> 
> I have mentioned separating SV from SNV, while optimizing vep.rule, to be able to allocate different cluster resources for SV and SNV, which is added by splitting the rule along with disabling vcfanno for SV.

Could you please elaborate on handling required vep results?

BALSAMIC/snakemake_rules/annotation/vep.rule
```diff
@@ -56,6 +54,53 @@ bcftools stats {output.vcf_all} > {output.bcftools_stats};
...
+export PERL5LIB=;
+
+bcftools reheader --threads {threads} -s {input.header} {input.vcf} \
+| bcftools view --threads {threads} -O z -o $tmpvcf ;
+
+vep \
+--dir $vep_path \
+--dir_cache {params.vep_cache} \
+--dir_plugins $vep_path \
+--input_file $tmpvcf \
```
hassanfa, line 92: Now that you've removed vcfanno, VEP can read STDIN. Probably just pipe it and remove tmpvcf.

BALSAMIC/snakemake_rules/annotation/vep.rule
```diff
@@ -56,6 +54,53 @@ bcftools stats {output.vcf_all} > {output.bcftools_stats};
...
+--output_file {output.vcf_all} \
+--fork {threads} \
+{params.vep_defaults} \
+--custom {input.cosmic},COSMIC,vcf,exact,0,CDS,GENE,STRAND,CNT,AA ;
+
+tabix -p vcf -f {output.vcf_all};
+
+bcftools stats {output.vcf_all} > {output.bcftools_stats};
+
+rm $tmpvcf;
```
hassanfa, line 102: remember to remove this if you use pipe.

hassanfa (commented):
👍🏽

hassanfa:
@khurrammaqbool Let me know on Slack again if you need me to have a look after the changes.

ashwini06:
> I have mentioned separating SV from SNV, while optimizing vep.rule, to be able to allocate different cluster resources for SV and SNV, which is added by splitting the rule along with disabling vcfanno for SV. It may be better to have separate rule with own allocated cluster resourced and it will also help monitoring it separately as well.

True, the current implementation looks good then. But I was thinking of an alternative solution, instead of having two rules based on var_type (SNV/SV), can we have separate rules based on rule_function (rule_vcfanno, rule_vep). This way in future,  if we wanna limit vcfanno functionality for certain callers we can do so. You can decide on the solution you wanna comeup with :)  And resource-wise, based on your previous investigation, vcfanno requires more compute time and threads right, VEP can fewer resources right? so can we reduce your current allocation for vep in cluster.json? 

> Could you please elaborate on handling required vep results?

Jus wondering if these files are generated correctly "SV.somatic.{case_name}.svdb.all.vcf.gz? Can you just make sure with some test runs..

ashwini06:
@khurrammaqbool Can you also check on the drop on the code coverage part?

BALSAMIC/snakemake_rules/annotation/vep.rule
khurrammaqbool: removed!

BALSAMIC/snakemake_rules/annotation/vep.rule
```diff
@@ -56,6 +54,53 @@ bcftools stats {output.vcf_all} > {output.bcftools_stats};
 rm $tmpvcf;
     """
 
+rule vep_somatic_sv:
+  input:
+    vcf = vcf_dir + "SV.somatic.{case_name}.svdb.vcf.gz",
+    header = vcf_dir + "SV.somatic.{case_name}.svdb.sample_name_map",
+    cosmic = config["reference"]["cosmic"]
```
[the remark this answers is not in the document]
khurrammaqbool: removed!

khurrammaqbool:
> > I have mentioned separating SV from SNV, while optimizing vep.rule, to be able to allocate different cluster resources for SV and SNV, which is added by splitting the rule along with disabling vcfanno for SV. It may be better to have separate rule with own allocated cluster resourced and it will also help monitoring it separately as well.
> 
> True, the current implementation looks good then. But I was thinking of an alternative solution, instead of having two rules based on var_type (SNV/SV), can we have separate rules based on rule_function (rule_vcfanno, rule_vep). This way in future, if we wanna limit vcfanno functionality for certain callers we can do so. You can decide on the solution you wanna comeup with :) And resource-wise, based on your previous investigation, vcfanno requires more compute time and threads right, VEP can fewer resources right? so can we reduce your current allocation for vep in cluster.json?

The current state of new SV annotation rule shows that it diverged from CNV annotation rule. It may continue to do so because obviously, it is not suitable to handle SNVs and SVs the same way. What you are suggesting now, is to further split SNV annotation into vep and vcfannno, which we can address this later after looking into it: as if we want to proceed with it or not. The handling of SNVs is, clearly, not scope of this PR and in fact separating will make it easy to take decisions and implement them on SNV and SV independently to be able to have variant type specific solutions, which is extremely hard when we generalize all variant types and try to handle them the same way. 

> > Could you please elaborate on handling required vep results?
> 
> Jus wondering if these files are generated correctly "SV.somatic.{case_name}.svdb.all.vcf.gz? Can you just make sure with some test runs..

Yes I could generate the files in test runs.

khurrammaqbool:
> @khurrammaqbool Can you also check on the drop on the code coverage part?

The drop in coverage is due to removal of annotation rules from delivery because we have decided that these generate intermediate files. The particular part effected is linked to housekeeper_id, which can be handled in the PR handling the tags and delivery of files.

khurrammaqbool:
> > @khurrammaqbool Can you also check on the drop on the code coverage part?
> 
> The drop in coverage is due to removal of annotation rules from delivery because we have decided that these generate intermediate files. The particular part effected is linked to housekeeper_id, which can be handled in the PR handling the tags and delivery of files.

Fixed!

ashwini06 (commented):
Looks good! For faster reviews, it will be nice if you let us know when PR is fixed and needs revision by tagging the people who need to approve the PR :)

BALSAMIC/snakemake_rules/annotation/vep.rule
```diff
@@ -56,6 +55,50 @@ bcftools stats {output.vcf_all} > {output.bcftools_stats};
 rm $tmpvcf;
     """
 
+rule vep_somatic_sv:
+  input:
+    vcf = vcf_dir + "SV.somatic.{case_name}.svdb.vcf.gz",
+    header = vcf_dir + "SV.somatic.{case_name}.svdb.sample_name_map",
+  output:
+    vcf_all = temp(vep_dir + "SV.somatic.{case_name}.svdb.all.vcf.gz"),
+    vcf_summary = vep_dir + "SV.somatic.{case_name}.svdb.all.vcf.gz_summary.html",
```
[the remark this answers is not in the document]
khurrammaqbool: This applies to all .all.vcf.gz, which are delivered at this point. To avoid confusion, the structure is maintained and will be fixed in PRs handling delivery files.

khurrammaqbool:
@ashwini06, please take a final look.

ashwini06 (approved):
👍🏼

PR #907, opened 2022-04-07
feat: extract additional WGS metrics

ivadym:
### This PR:

Added new WGS metrics to be extracted and stored to `<case>_metrics_deliverables.yaml`.

### Review and tests: 
- [ ] Tests pass
- [ ] Code review

### DIFF
--- a/BALSAMIC/constants/quality_check_reporting.py
+++ b/BALSAMIC/constants/quality_check_reporting.py
@@ -94,7 +94,6 @@ METRICS = {
         "MEAN_INSERT_SIZE": {"condition": None},
         "MEDIAN_COVERAGE": {"condition": None},
         "FastQC_mqc-generalstats-fastqc-percent_duplicates": {"condition": None},
-        "PCT_5X": {"condition": None},
         "PCT_15X": {"condition": None},
         "PCT_30X": {"condition": None},
         "PCT_60X": {"condition": None},
--- a/BALSAMIC/constants/quality_check_reporting.py
+++ b/BALSAMIC/constants/quality_check_reporting.py
@@ -90,5 +90,15 @@ METRICS = {
             "FOLD_80_BASE_PENALTY": {"condition": {"norm": "lt", "threshold": 1.8}},
         },
     },
-    "wgs": {"FOLD_80_BASE_PENALTY": {"condition": {"norm": "lt", "threshold": 1.8}}},
+    "wgs": {
+        "MEAN_INSERT_SIZE": {"condition": None},
+        "MEDIAN_COVERAGE": {"condition": None},
+        "FastQC_mqc-generalstats-fastqc-percent_duplicates": {"condition": None},
+        "PCT_5X": {"condition": None},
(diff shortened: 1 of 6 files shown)


ashwini06 (approved):
Looks good 💯

BALSAMIC/constants/quality_check_reporting.py
```diff
@@ -90,5 +90,15 @@
             "FOLD_80_BASE_PENALTY": {"condition": {"norm": "lt", "threshold": 1.8}},
         },
     },
-    "wgs": {"FOLD_80_BASE_PENALTY": {"condition": {"norm": "lt", "threshold": 1.8}}},
+    "wgs": {
+        "MEAN_INSERT_SIZE": {"condition": None},
+        "MEDIAN_COVERAGE": {"condition": None},
+        "FastQC_mqc-generalstats-fastqc-percent_duplicates": {"condition": None},
+        "PCT_5X": {"condition": None},
```
keyvanelhami, line 97: Have we agreed on having `PCT_5X` for WGS? Or did you just add it in case we would like to have this criteria in the future? Same thing for `PCT_100X`

BALSAMIC/constants/quality_check_reporting.py
```diff
@@ -90,5 +90,15 @@
...
+        "PCT_15X": {"condition": None},
+        "PCT_30X": {"condition": None},
+        "PCT_60X": {"condition": None},
+        "PCT_100X": {"condition": None},
+        "FOLD_80_BASE_PENALTY": {"condition": {"norm": "lt", "threshold": 1.8}},
```
keyvanelhami, line 102: So the fold-80 will only be used for the delivery report, correct? I.e. not for failing an analysis if the value is higher than 1.8?

BALSAMIC/constants/quality_check_reporting.py
ivadym: @keyvanelhami: `15X` and `60X` are used for the delivery report. The rest of the PCT values are retrieved simply by following the progression of the already defined target values, so there is not much behind their choice. If you want to suggest another (and probably more relevant set of PCT values) please do it.

BALSAMIC/constants/quality_check_reporting.py
ivadym: I was only addressing the retrieval of values for the delivery report since the discussion about the QC criteria is still open.

But i can remove `FOLD_80` and `MEDIAN_COVERAGE` values as a failing criteria for both `WGS` and `PANEL`, although I'm not sure about the remaining metrics and their conditions. Have you agreed on something specific?

Thanks!

keyvanelhami: Lets then keep as it is today until we agree on what QC criteria to use

BALSAMIC/constants/quality_check_reporting.py
keyvanelhami: You can remove `PCT_5X` but lets keep the rest

Issue #908, opened 2022-04-07
Production runs fails after v8.2.9 deployment


ashwini06:
**Describe the bug**
After deployment of v8.2.9 production runs fails at `vep_somatic` and `vep_germline` rules

<img width="1440" alt="Screenshot 2022-04-07 at 11 06 39" src="https://user-images.githubusercontent.com/6302819/162163565-ebd95c21-a39e-4a17-9e26-9269239bf7ce.png">

log from : `/home/proj/production/cancer/cases/raredragon/logs/BALSAMIC.raredragon.vep_somatic.84.sh_2811378.err`

**Problem**: bcftools is not installed in annotate container 
```Activating singularity image /home/proj/production/cancer/balsamic_cache/8.2.9/containers/annotate.sif
/.singularity.d/actions/exec: /.singularity.d/env/10-docker2singularity.sh: line 1: ash\: not found
bcftools: error while loading shared libraries: libopenblas.so.0: cannot open shared object file: No such file or directory
bcftools: error while loading shared libraries: libopenblas.so.0: cannot open shared object file: No such file or directory

=================
```

**Version (please complete the following information):**
`balsamic v8.2.9` 

**Solution**
Currently bcftools version is unlocked and bioconda searches for greater version than >1.10 which might need openssl or other depencies.

Lock the bcftools version to `1.10` in https://github.com/Clinical-Genomics/BALSAMIC/blob/master/BALSAMIC/containers/annotate/annotate.yaml for quick fix

PR #869, opened 2022-02-24
feat: add docker container for vcf2cytosure

ashwini06:
### This PR:

Previously `vcf2cytosure` is added to the annotate container in PR #650. But there exists a problem loading `vcf2cytosure` in this container

aa1

The problem might be with the loading entry points for vcf2cytosure installation and its installation exited..

Removed: vcf2cytosure from `annotate` container

Added: New Docker container for vcf2cytosure. Related to #858 

### Review and tests: 
- [ ] Tests pass
- [ ] Code review
- [ ] New code is executed and covered by tests, and test approve

PR #909, opened 2022-04-07
fix: bcftools locked in annotate container

ashwini06:
### This PR:

Fixes : #908 

### Review and tests: 
- [ ] Tests pass
- [ ] Code review

### DIFF
--- a/BALSAMIC/containers/annotate/annotate.yaml
+++ b/BALSAMIC/containers/annotate/annotate.yaml
@@ -6,7 +6,8 @@ channels:
 dependencies:
   - anaconda::python=3.7
   - bioconda::ensembl-vep=100.2
-  - bioconda::bcftools>=1.10
+  - bioconda::bcftools==1.10
+  - bioconda::openssl=1.0
   - bioconda::vcfanno=0.3.3
   - anaconda::gxx_linux-64=7.3.0
   - anaconda::pip=20.2.4
--- a/BALSAMIC/containers/annotate/annotate.yaml
+++ b/BALSAMIC/containers/annotate/annotate.yaml
@@ -7,7 +7,6 @@ dependencies:
   - anaconda::python=3.7
   - bioconda::ensembl-vep=100.2
   - bioconda::bcftools==1.10
-  - bioconda::openssl=1.0
   - bioconda::vcfanno=0.3.3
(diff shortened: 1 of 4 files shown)


BALSAMIC/containers/annotate/annotate.yaml
```diff
@@ -6,7 +6,7 @@ channels:
 dependencies:
   - anaconda::python=3.7
   - bioconda::ensembl-vep=100.2
-  - bioconda::bcftools>=1.10
+  - bioconda::bcftools==1.10
```
khurrammaqbool, line 9: - bioconda::bcftools=1.10

khurrammaqbool (approved):

container_tests/annotate/annotate.sh
```diff
@@ -1,15 +1,16 @@
...
-  fi
+    tool_msg=$($tool --help 2>&1 | grep 'No')
+    if [[ -z $tool_msg ]]; then 
+        echo "$tool command exists and works";
+    else 
+        echo "$tool command doesnt work";
+        exit 1;
+    fi
 done
+
```
ivadym, line 16: This is not working as expected: since you are checking if the `tool_msg` is of length 0, all the `valid_commands` (even the ones with arbitrary names that does not exist in the binaries folder) will be evaluated as True with this parsing `| grep 'No'`.

Screen Shot 2022-04-11 at 10 28 39

Maybe this if statement could be combined with the previous implementation but I'm not sure if there is a good way to tests for dependencies as it could be hard to generalise it for all packages.

ivadym (changes requested):

container_tests/annotate/annotate.sh
ashwini06: Thanks for the suggestion @ivadym :) Update with new test

Screenshot 2022-04-11 at 12 26 27

ivadym (approved):
No further comments from my side 👍

ashwini06:
Drop-in code coverage is due to the removal of pip installation of vcf2cytosure , which would remove coverage for this line https://app.codecov.io/gh/Clinical-Genomics/BALSAMIC/compare/909/changes#D1L475. But this is already addressed in develop branch in PR https://github.com/Clinical-Genomics/BALSAMIC/pull/869. Ignoring the drop and merging.

PR #919, opened 2022-04-22
fix: Change of gnomad pop freq value for UMI workflow

ashwini06:
### This PR:

Changed: 

- For UMI workflow, post-filter `gnomad_pop_freq` value is changed from `0.005` to `0.02`. This is based on the suggestion from Emma Tham.
- Modified the readthedocs for balsamic variant-calling filters

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


docs/balsamic_filters.rst
```diff
@@ -265,10 +293,25 @@ Minimum log-odds for the candidate selection. TNscope default: `4`. In our UMI-w
...
 *GNOMADAF_POPMAX*: Maximum Allele Frequency across populations
 
 ::
 
-    GNOMADAF_popmax <= 0.001 (or) GNOMADAF_popmax == "."
+    GNOMADAF_popmax <= 0.02 (or) GNOMADAF_popmax == "."
+
+.. attention::
+   BALSAMIC <= v8.2.10 uses GNOMAD_popmax <= 0.005. From Balsamic v9.0.0, this settings is changed to 0.02
+
```
khurrammaqbool, line 317: Please add, how will it effect the results in terms of number of variants for clarity.

BALSAMIC/constants/variant_filters.py
```diff
@@ -47,6 +47,11 @@
         "filter_name": "balsamic_high_pop_freq",
         "field": "INFO",
     },
+    "pop_freq_umi": {
+        "tag_value": 0.02,
+        "filter_name": "balsamic_umi_high_pop_freq",
```
khurrammaqbool, line 52: balsamic_umi_max_pop_freq

BALSAMIC/constants/variant_filters.py
```diff
@@ -47,6 +47,11 @@
...
+        "field": "INFO",
+    },
     "qss": {
         "tag_value": 20,
         "filter_name": "balsamic_low_quality_scores",
```
khurrammaqbool, line 57: balsamic_low_quality_variant_scores

BALSAMIC/utils/models.py
```diff
@@ -56,6 +56,7 @@ class VarCallerFilter(BaseModel):
         MQ: VCFAttributes (optional); minimum mapping quality
         DP: VCFAttributes (optional); minimum read depth
         pop_freq: VCFAttributes (optional); maximum gnomad_af
+        pop_freq_umi: VCFAttributes (optional); maximum gnomad_af for UMI workflow
```
khurrammaqbool, line 59: max_pop_freq_umi

BALSAMIC/utils/models.py
```diff
@@ -71,6 +72,7 @@ class VarCallerFilter(BaseModel):
     MQ: Optional[VCFAttributes]
     DP: Optional[VCFAttributes]
     pop_freq: Optional[VCFAttributes]
+    pop_freq_umi: Optional[VCFAttributes]
```
khurrammaqbool, line 75: max_pop_freq_umi

docs/balsamic_filters.rst
```diff
@@ -37,7 +37,35 @@ indicate that at this site, the mean position in reads is less than 8, and the p
...
+Only those variants that fulfill the pre-call and post-call filters are scored as `PASS` in the `STATUS` column of the VCF file. We filter those `PASS` variants and deliver a final list of variants to the customer either via `Scout` or `Caesar`
+
+.. note::
+    In BALSAMIC, this VCF file is named as `*.all.filtered.pass.vcf.gz` (eg: `SNV.somatic.<CASE_ID>.vardict.all.filtered.pass.vcf.gz`) 
+
+
+.. list-table:: Description of VCF files
+    :widths: 30 50 20
+    :header-rows: 1
```
khurrammaqbool, line 51: This appears above the table after rendering.

BALSAMIC/constants/variant_filters.py
ashwini06: @khurrammaqbool : The naming is inline according to the previously defined names in https://github.com/Clinical-Genomics/BALSAMIC/blob/af98af67d6071a3ed8584da5e0e08ff6f145b2a4/BALSAMIC/constants/variant_filters.py

If you suggest the `max` instead of `high`, then we need to agree to change it across all in https://github.com/Clinical-Genomics/BALSAMIC/blob/af98af67d6071a3ed8584da5e0e08ff6f145b2a4/BALSAMIC/constants/variant_filters.py
and also relevant changes in snakemake rules

docs/balsamic_filters.rst
ashwini06: It might increase the number of variants by lowering this threshold and no specific test validations are done to check the increased number.

ashwini06:
> I noticed after review that the PR is draft. Please ignore those comments that are already fixed!

@khurrammaqbool @keyvanelhami : Ready for review

docs/balsamic_filters.rst
khurrammaqbool: We can add the effect to the text and there is no need to provide the numbers as it may vary from case to case.

docs/balsamic_filters.rst
```diff
@@ -265,10 +293,25 @@ Minimum log-odds for the candidate selection. TNscope default: `4`. In our UMI-w
 
     min_tumor_lod = 4.0
 
+*min_tumor_allele_frac*: Set the minimum tumor AF to be considered as potential variant site.
+
+::
+
+    min_tumor_allele_frac = 0.0005
+
+*interval_padding*:  Amount to pad all intervals in the given target bed file regions
```
keyvanelhami, line 302: Please clarify what padding means since this is unclear for the customer:
"Adding an extra 100bp to each end of the target region before variant calling"

keyvanelhami (approved):

docs/balsamic_filters.rst
ashwini06: Fixed!

docs/balsamic_filters.rst
ashwini06: Fixed!

BALSAMIC/constants/variant_filters.py
ashwini06: Leaving it as such to be in inline with previous names, as it includes much more modifications.

PR #969, opened 2022-06-23
feat: add tabix and gzip to vcf2cytosure container

khurrammaqbool:
### This PR:

Added: tabix and gzip to vcf2cytosure container

### Review and tests: 
- [ ] Tests pass
- [ ] Code review

### DIFF
--- a/BALSAMIC/containers/vcf2cytosure/vcf2cytosure.yaml
+++ b/BALSAMIC/containers/vcf2cytosure/vcf2cytosure.yaml
@@ -1,1 +1,2 @@
 - vcf2cytosure=0.7.1
+- tabix=1.11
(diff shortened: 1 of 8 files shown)


BALSAMIC/containers/vcf2cytosure/vcf2cytosure.yaml
```diff
@@ -1 +1,2 @@
 - vcf2cytosure=0.7.1
+- tabix=1.11
```
ashwini06, line 2: Purpose of having tabix here and also in `vcf2cytosure.yaml` ?

ashwini06 (commented):
Adding tabix to the vcf2cytosure, will fix bgzip but not zgrep. We have both bcftools and tabix installed (for eg: in `annotate` container), but still `zgrep` doesn't work in it. 

I think you need to add the `gzip` package to the Dockerfile for fixing zgrep.

PR #973, opened 2022-06-29
fix: command and condition for tiddit and fixed ReadtheDocs

khurrammaqbool:
### This PR:

Fixed: Command and condition for TIDDIT and fixed ReadtheDocs

### Review and tests: 
- [ ] Tests pass
- [ ] Code review

### DIFF
--- a/BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_normal.rule
+++ b/BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_normal.rule
@@ -237,7 +237,7 @@ rule ascat_tumor_normal_merge_output:
 python {params.merge_ascat_output_script} {output.ascat_pdf} {input.sample_statistics} {input.ascat_plots}
         """
 
-rule tiddit_sv_tumor_normal:
+rule tiddit_tumor_normal:
     input:
         fa = config["reference"]["reference_genome"],
         bamN = bam_dir + normal_bam,
@@ -248,7 +248,7 @@ rule tiddit_sv_tumor_normal:
         cov_normal_tiddit = vcf_dir + "SV.somatic." + config["analysis"]["case_id"] + ".normal.tiddit_cov.bed",
         namemap= vcf_dir + "SV.somatic." + config["analysis"]["case_id"] + ".tiddit.sample_name_map",
     benchmark:
-        benchmark_dir + 'tiddit_sv_tumor_normal_' + config["analysis"]["case_id"] + ".tsv"
+        benchmark_dir + 'tiddit_tumor_normal_' + config["analysis"]["case_id"] + ".tsv"
     singularity:
         Path(singularity_image, config["bioinfo_tools"].get("tiddit") + ".sif").as_posix()
     params:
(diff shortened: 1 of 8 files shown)


CHANGELOG.rst
```diff
@@ -41,6 +41,7 @@ Fixed:
 * Command in vcf2cytosure rule and updated ReadtheDocs https://github.com/Clinical-Genomics/BALSAMIC/pull/966
 * Missing name `analysis_dir` in QC.smk https://github.com/Clinical-Genomics/BALSAMIC/pull/970
 * Remove `sample_type` wildcard from the `vcfheader_rename_germline` rule and change genotype file name https://github.com/Clinical-Genomics/BALSAMIC/pull/971
+* TIDDIT command and a typographical error in ReadtheDocs https://github.com/Clinical-Genomics/BALSAMIC/pull/973
```
ashwini06, line 44: Will this change gonna be a patch fix 10.0.1?

ashwini06 (approved):
👍🏼

CHANGELOG.rst
```diff
@@ -41,6 +41,7 @@ Fixed:
 * Command in vcf2cytosure rule and updated ReadtheDocs https://github.com/Clinical-Genomics/BALSAMIC/pull/966
 * Missing name `analysis_dir` in QC.smk https://github.com/Clinical-Genomics/BALSAMIC/pull/970
 * Remove `sample_type` wildcard from the `vcfheader_rename_germline` rule and change genotype file name https://github.com/Clinical-Genomics/BALSAMIC/pull/971
+* Command and condition for TIDDIT and fixed ReadtheDocs https://github.com/Clinical-Genomics/BALSAMIC/pull/973
```
ashwini06, line 44: Can you move this to [10.0.1]

BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_normal.rule
```diff
@@ -258,7 +258,7 @@ rule tiddit_sv_tumor_normal:
         normal = "NORMAL",
         case_name = config["analysis"]["case_id"]
     threads:
-        get_threads(cluster_config, "tiddit_sv_tumor_normal")
+        get_threads(cluster_config, "tiddit_tumor_normal")
```
ashwini06, line 261: This name differs from the rulename. If you would like to use this name instead, it's good to change the rule name accordingly and also other related names within this rule.

khurrammaqbool: I suggest that the cluster resources allocation should be method dependent and not rule dependent. It will enable us to use common resources for multiple rules and avoid an unnecessary redundant declaration of the resources for the same method. I was thinking to add major variant type in the rule names to make it coherent across variant calling rules when we plan for refactoring anytime after the current release. It would be nice to take the point of view from @rannick and @ivadym about it.

ivadym: If we can avoid code repetition in `cluster.json` and simplify it, I think it's a good idea to group the resource definition as long as no extra resources are allocated because of this. We should not allocate more resources than are strictly needed.

ashwini06: OK, not exactly sure of your future plans on this. Then it includes modifying naming across all rules and it deviates from the current standard structure we defined for our snakemake rules here https://balsamic.readthedocs.io/en/latest/snakemake_etiquette.html#structure-of-snakemake-rules

ivadym: Actually the real problem here is having multiple rules that are almost identical, just with the difference of handling tumor or tumor and normal samples. Having one or two more lines in cluster.json does not bring anything when we have two rules that are practically the same...

PR #1012, opened 2022-10-18
feat: add clinical observations and refactor toml

khurrammaqbool:
### This PR:

Added: add clinical observations and toml configuration method for vcfanno.
Changed: refactor vcfanno toml config.

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


BALSAMIC/workflows/balsamic.smk
```diff
@@ -58,9 +59,35 @@ qc_dir = get_result_dir(config) + "/qc/"
...
+    clinical_observations = get_annotation(
+        clinical_observations_path,
+        ["Frq", "Obs", "Hom"],
+        ["self", "self", "self"],
+        ["Frq", "Obs", "Hom"]
+    )
+    annotations = [gnomad, clinvar, clinical_observations]
+else:
+    clinical_observations = None
+    annotations = [gnomad, clinvar]
```
ivadym, line 88: I would recommend to have this as a constants and remove the `get_annotation` method.

ivadym (commented):
Well done! 🌟  

Left you some comments @khurrammaqbool, tell me what you think :)
I'm also missing some typehints and docstrings in the methods in `rule.py`

khurrammaqbool:
> Well done! 🌟
> 
> Left you some comments @khurrammaqbool, tell me what you think :) I'm also missing some typehints and docstrings in the methods in `rule.py`

Thanks for reminding me!

BALSAMIC/utils/rule.py
```diff
@@ -350,3 +351,45 @@ def get_pon_samples(fastq_dir):
...
+
+    Args:
+        congig: a config dictionary
+
+    Returns:
+        Path for clinical_snv_observations vcf file
+
+    """
+    if "clinical_snv_observations" in config["reference"]:
+        return Path(config["reference"]["clinical_snv_observations"]).as_posix()
```
ivadym, line 367: What should be returned in case there are no clinical observations? A `None`? an empty string? And how this empty field will be treated later by the rules?

BALSAMIC/utils/rule.py
```diff
@@ -350,3 +351,45 @@ def get_pon_samples(fastq_dir):
...
+    """Returns annotation in toml format
+
+    Args:
+        annotation: a dict containing annotation resource
+
+    Returns:
+        toml_annotation: a string in toml format
+    """
+    toml_annotation = toml.dumps(annotation)
+    return toml_annotation
```
ivadym, line 380:
```suggestion
    return toml.dumps(annotation)
```

BALSAMIC/utils/rule.py
```diff
@@ -350,3 +351,45 @@ def get_pon_samples(fastq_dir):
...
+
+def dump_toml(annotations: list) -> str:
+    """Returns list of converted annotation in toml format
+
+    Args:
+        annotations: a list of toml
+
+    Returns:
+        toml_annotation: a string of toml annotation resources
+    """
```
ivadym, line 391: Nice ⭐

BALSAMIC/utils/rule.py
```diff
@@ -350,3 +351,45 @@ def get_pon_samples(fastq_dir):
...
+    Returns:
+        Path for clinical_snv_observations vcf file
+
+    """
+    if "clinical_snv_observations" in config["reference"]:
+        return Path(config["reference"]["clinical_snv_observations"]).as_posix()
+
+
+def get_toml(annotation: dict) -> str:
+    """Returns annotation in toml format
```
ivadym, line 371: We should add some `pytest` for these methods. Some time ago I created an `io` file (in utils directory), do you think the `toml` methods would fit better there?

ivadym (commented):
Looks much better now, just left you some minor comments :)

BALSAMIC/utils/rule.py
khurrammaqbool: As we discussed, since we are using a python package, which is already well covered by the tests, it may not be required here to be redone.

BALSAMIC/utils/rule.py
ivadym: This method does not return `None`, so we have to provide a return statement for all of the possible cases.

So I would go for one of the following solutions, since we already have an `if` statement in `balsamic.smk`:

```suggestion
    return Path(config["reference"]["clinical_snv_observations"]).as_posix()
```
```suggestion
    if "clinical_snv_observations" in config["reference"]:
        return Path(config["reference"]["clinical_snv_observations"]).as_posix()
    return None
```

BALSAMIC/utils/rule.py
ivadym: I think at least for `dump_toml` i would be useful to test the addition assignment and that the output structure corresponds to what you expect

ivadym (approved):
Fantastic! 

Leaving the refactoring for future pushes as we discussed 👍

PR #1013, opened 2022-10-20
feat: clinical sv annotation and SV SNV filter rules

khurrammaqbool:
### This PR:

Added: Clinical SV annotation for somatic SVs and SV,SNV filter rules.

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


ivadym (approved):
Great job! 👍 

We would also need to update the input of the `collect_custom_qc_metrics` rule to take the clinical files, but I can do that in another PR after you restructure the annotation & filtering.

PR #1025, opened 2022-11-25
fix: invalid flag for mate and update commands

khurrammaqbool:
### This PR:

Fixed: invalid flag for mate and update commands

### Review and tests: 
- [ ] Tests pass
- [ ] Code review
- [ ] New code is executed and covered by tests, and test approve

PR #1026, opened 2022-11-26
fix: Revert "fix: invalid flag for mate and update commands"

ivadym:
Reverts Clinical-Genomics/BALSAMIC#1025


ivadym:
Mistakes are made 🤦‍♂️

Issue #1061, opened 2023-01-18
Explicitly specify versions for all dependencies for containers


fevac:
### Description

Currently containers often fails when building due to changes in versions of unspecified requirements (e.g. tools dependencies). For reproducibility purposes it would be nice to set all versions of all tools in the environment. Once environments are fully specified, it would be good to avoid build testing on every single push as it often takes a long time. 

## Suggested solution

- Replace requirement .yaml files with the output of `conda env export` of working containers
- Add `paths` on the docker build actions it only runs when files within the `containers` directories are modified

**BALSAMIC version**
`11.0.2`

ivadym:
Addressed in: https://github.com/Clinical-Genomics/BALSAMIC/pull/1096.

@Clinical-Genomics/cancer-bioinfo: FYI I'll pull the updated containers from docker hub and add them to the 11.2.0 cache in stage

PR #1062, opened 2023-01-19
build: Update env requirements for containers

fevac:
### This PR:

Updates containers specifications by explicitly setting versions for all dependencies. 

Related to: https://github.com/Clinical-Genomics/BALSAMIC/issues/1061

**Note:**
- Requirements for balsamic container might be problematic since the balsamic version cannot be specified as it will be changing. If this is actually the case and errors often arise, we might want to revert the specifications in this container back to the minimal settings. 

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


fevac:
**NOTE**

When exporting an environment with `conda env export` it adds a platform-specific version tag. This will be problematic when switching between platforms (eg. mac and linux). To avoid including the build tags in conda environment files we should export using `conda env export --no-builds` instead.

ivadym:
It will also be nice to see and compare how long it takes to install them and if we see an improvement there as well

fevac:
I'm officially giving up on this pytest errors (which I also get locally and I don't think they are container related??). I'll put this on hold for now and check further once I have more time again.

fevac:
Closing in favour of https://github.com/Clinical-Genomics/BALSAMIC/pull/1096

PR #1074, opened 2023-01-25
New template for internal issues

pbiology:
### This PR:

This just adds a new issue template for use for internal development. The existing ones didn't feel geared towards this.
A question for those who review: Do we use the "Custom issue template" for any new issues? It feels a bit out of place and we could also remove it in this PR if it isn't needed.


ivadym (approved):
👍

PR #1080, opened 2023-01-31
fix: change somalier container user to defaultuser

fevac:
### This PR:

Addresses security hotspot related to using alpine docker with the default root user in somalier. It changes it to `defaultuser` instead

Fixed: somalier container user to non-root user

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


ivadym (approved):
Well done 🤩  We'll have to make these changes for all the containers though (maybe when we start splitting them)

But let's hope we don't have any more unexpected github actions complains... 🥵

PR #1082, opened 2023-02-01
feat: Create add-to-bl.yml

pbiology:
### This PR:
- Adds a GitHub action which will add all new issues to the Cancer BL project

In the future this could be tweaked so only issues marked with certain labels etc get's added, but I think we can start with all issues

They will be added to the "To Discuss" column.

### DIFF
--- a/.github/workflows/add-to-bl.yml
+++ b/.github/workflows/add-to-bl.yml
@@ -1,0 +1,17 @@
+name: Add new issues to the backlog
+
+on:
+  issues:
+    types:
+      - opened
+
+jobs:
+  add-to-project:
+    name: Add issue to project
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/add-to-project@v0.4.0
+        with:
+          project-url: https://github.com/orgs/Clinical-Genomics/projects/45
+          github-token: ${{ secrets.ADD_TO_PROJECT_PAT }}
+          
--- a/.github/workflows/add-to-bl.yml
+++ b/.github/workflows/add-to-bl.yml
@@ -13,4 +13,4 @@ jobs:
       - uses: actions/add-to-project@v0.4.0
         with:
           project-url: https://github.com/orgs/Clinical-Genomics/projects/45
-          github-token: ${{ secrets.ADD_TO_PROJECT_PAT }}
+          github-token: ${{ secrets.REPO_AND_PROJECT_PAT }}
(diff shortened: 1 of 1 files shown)


.github/workflows/add-to-bl.yml
```diff
@@ -0,0 +1,17 @@
...
+jobs:
+  add-to-project:
+    name: Add issue to project
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/add-to-project@v0.4.0
+        with:
+          project-url: https://github.com/orgs/Clinical-Genomics/projects/45
+          github-token: ${{ secrets.ADD_TO_PROJECT_PAT }}
+
```
ivadym, line 17: Trailing space

```suggestion

```

.github/workflows/add-to-bl.yml
```diff
@@ -0,0 +1,17 @@
...
+jobs:
+  add-to-project:
+    name: Add issue to project
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/add-to-project@v0.4.0
+        with:
+          project-url: https://github.com/orgs/Clinical-Genomics/projects/45
+          github-token: ${{ secrets.ADD_TO_PROJECT_PAT }}
```
ivadym, line 16: It's probably a better idea to use the [automatic token authentication](https://docs.github.com/en/actions/security-guides/automatic-token-authentication) that having an empty secret, what do you think?

```suggestion
          github-token: ${{ secrets.GITHUB_TOKEN }}
```

.github/workflows/add-to-bl.yml
```diff
@@ -0,0 +1,17 @@
...
+
+on:
+  issues:
+    types:
+      - opened
+
+jobs:
+  add-to-project:
+    name: Add issue to project
+    runs-on: ubuntu-latest
```
ivadym, line 11: We have been fixing it to a specific version. I don't have a strong opinion on this (in cg we have ubuntu-latest for example), but we should homogenise it here

```suggestion
    runs-on: ubuntu-20.04
```

ivadym (commented):
Nice 🔥

The Conventional PR GitHub Action is failing because we are using [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) types for PR titles (`feat`, `fix:`, `perf:`, etc).

.github/workflows/add-to-bl.yml
pbiology: Good suggestions

.github/workflows/add-to-bl.yml
pbiology: I've added an org PAT now. Should work. It will expire at the end of the year, so if this breaks that's why :)

ivadym (approved):
Great feature! 👍

PR #1084, opened 2023-02-03
chore: change git action to build containers only when they have been modified

fevac:
### This PR: 

Changed: git actions to build docker containers only if the `containers` directory is modified.

Related to: https://github.com/Clinical-Genomics/BALSAMIC/issues/1061

### Review and tests: 
- [ ] Tests pass
- [ ] Code review

### DIFF
--- a/.github/workflows/docker_build_push.yml
+++ b/.github/workflows/docker_build_push.yml
@@ -4,6 +4,8 @@ on:
     branches:
       - master
       - develop
+    paths:
+      - BALSAMIC/containers/
 jobs:
   main:
     name: Docker image build push
--- a/.github/workflows/add-to-bl.yml
+++ b/.github/workflows/add-to-bl.yml
@@ -1,0 +1,16 @@
+name: Add new issues to the backlog
+
+on:
+  issues:
+    types:
+      - opened
+
+jobs:
+  add-to-project:
+    name: Add issue to project
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/add-to-project@v0.4.0
+        with:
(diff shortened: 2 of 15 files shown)


fevac:
@ivadym maybe we want to keep the building and testing for the releases regardless of the containers have been modified or not? not sure what your thoughts are about this

ivadym:
Exactly, we have to keep testing, building and pushing them to docker hub for pushes to master and develop (this should be left as it is in my opinion). 

For the rest of the branches we ignore the test unless we make changes to the containers. It would be also nice to just build the specific containers that we modify, but I couldn't find a dynamic way to do it, i.e.:

```
on:
...
    paths:
      - BALSAMIC/containers/[container name]/
    ...
```

fevac:
Agree. It would be nice if it was container specific but we can see this as a partial improvement and change again in the future if we find a way.

ivadym (approved):
Looks good 💯 🚀

.github/workflows/docker_build_push.yml
```diff
@@ -4,6 +4,8 @@ on:
     branches:
       - master
       - develop
+    paths:
+      - BALSAMIC/containers/
```
ivadym, line 8: Just realised that the syntax may be incorrect, although I'm not sure if it will work with just BALSAMIC/containers/

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

```suggestion
      - "BALSAMIC/containers/**"
```

ivadym:
Have been addressed here: https://github.com/Clinical-Genomics/BALSAMIC/pull/1221

PR #1089, opened 2023-02-09
fix: Number of variants increased with inclusion of triallelic_sites

khurrammaqbool:
### This PR:

Fixed: Number of variants increased with inclusion of triallelic_sites

### Review and tests: 
- [ ] Tests pass
- [ ] Code review

### DIFF
--- a/BALSAMIC/snakemake_rules/annotation/varcaller_filter_tumor_normal.rule
+++ b/BALSAMIC/snakemake_rules/annotation/varcaller_filter_tumor_normal.rule
@@ -106,10 +106,10 @@ rule bcftools_filter_TNscope_umi_research_tumor_normal:
     "adding FOUND_IN tags to the output VCF file for {params.case_name} "
   shell:
     """
-bcftools view  -f PASS,triallelic_site --threads {threads} {input.vcf_snv_research} | \
+bcftools view --threads {threads} -f PASS,triallelic_site {input.vcf_snv_research} | \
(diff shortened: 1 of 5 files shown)


khurrammaqbool:
## civilsole (WGS-TO) 

## clinical filtered pass
   Number of samples: 1
   Number of SNPs:    29065
   Number of INDELs:  4366
   Number of MNPs:    0
   Number of others:  0
   Number of sites:   33431
## research filtered pass
  Number of samples: 1
  Number of SNPs:    29454
  Number of INDELs:  4902
  Number of MNPs:    0
  Number of others:  0
  Number of sites:   34356

fevac:
Great! ⭐  What filtering did you do at the end? I'm not as fluent with bcftools so I don't fully follow

khurrammaqbool:
> Great! ⭐ What filtering did you do at the end? I'm not as fluent with bcftools so I don't fully follow

I same filters, `PASS` and `triallelic_site`, are added as an expression. 

The text below is taken from the documentation of `bcftools` that explains it
`include only sites for which EXPRESSION is true`

The documentation of `bcftools` is a great resource to become "fluent" with it.
https://samtools.github.io/bcftools/bcftools.html

mathiasbio:
If I understood the changes of the filtering correctly this will mean that variants with either PASS or only the triallelic_site filter set will be kept in the final vcf. Using the -f flag instead would also keep variants with the triallelic_site filter even if they had in addition other filters set. This sounds like the fix implemented the original intended filtering strategy!

mathiasbio (approved):
Looks like the fix should work as intended! Nice job : )

PR #1093, opened 2023-02-13
fix: no more pytest generated files in my directory 

ivadym:
### This PR:

Removed:
* Pytest file saved in balsamic directory

### Review and tests: 
- [x] Tests pass
Screenshot 2023-02-13 at 17 48 42

- [ ] Code review


tests/commands/plugins/conftest.py
```diff
@@ -7,8 +7,8 @@ def input_file():
 
 
 @pytest.fixture
-def output_file():
-    return "tests/test_data/vcf_tables/test_createVCF_output.vcf.gz"
+def output_file(tmp_path):
+    return tmp_path / "test_createVCF_output.vcf.gz"
```
fevac, line 11:
```suggestion
    return Path(tmp_path, "test_createVCF_output.vcf.gz").as_posix()
```

fevac (changes requested):
🌟 🌟

tests/commands/plugins/conftest.py
ivadym: 🕺

fevac (approved):

PR #1096, opened 2023-02-22
refactor: Fix all dependencies to specific versions

ivadym:
### This PR:

- Fixed all conda container dependencies
- Add pip specific missing tools to config

Balsamic container not included as it depends on the release version.

### Review and tests: 
- [x] Tests pass

#### TN-WGS
Screenshot 2023-02-22 at 13 40 55

#### TN-TGA
Screenshot 2023-02-22 at 13 41 56

PIP specific dependencies (`cnvkit`) are included now in the `config.json` file
Screenshot 2023-02-22 at 13 43 30

- [ ] Code review


BALSAMIC/containers/vcf2cytosure/vcf2cytosure.yaml
```diff
@@ -1 +1,72 @@
...
+      - click==8.1.3
+      - coloredlogs==15.0.1
+      - cyvcf2==0.30.18
+      - humanfriendly==10.0
+      - lxml==4.9.2
+      - numpy==1.24.2
+      - pandas==1.5.3
+      - python-dateutil==2.8.2
+      - pytz==2022.7.1
+      - vcf2cytosure==0.7.1
```
ivadym, line 72: Probably because the last version (0.8) of vcf2cytosure is not published in pypi it appear as 0.7.1. However, 0.8 is the one that actually is installed (as stated in the vcf2cytosure Dockerfile)
Screenshot 2023-02-22 at 09 50 33

henrikstranneheim: @dnil do you have time to publish to pypi? One can also add an action to do it on release.

dnil: Sure! We have that for so many that I likely just forgot. 😊

I suppose the reason for no automation is that it is still under NBIS, so we did not have easy access to secrets etc.

dnil: ...its probably mostly the vcf2cytosure setup.py having its own version and not fetching a dynamic one.  I'll fix that first, and the version should look right for pip from GitHub as well. 😊

ivadym: That's true, it's taking that version from `setup.py`, which is `0.7.1`. Thanks @dnil!

dnil: Ok, thank you for noticing the problem! Give it a shot now: `0.8.1` should have version set dynamically!

BALSAMIC/containers/align_qc/align_qc.yaml
```diff
@@ -1,16 +1,194 @@
...
-  - bioconda::multiqc=1.12
-  - bioconda::fastp=0.23.2
-  - conda-forge::csvkit=1.0.7
-  - conda-forge::libiconv=1.16
-  - conda-forge::fontconfig=2.14.0
-  - conda-forge::r-base=4.1.3
+  - _libgcc_mutex=0.1=main
+  - _openmp_mutex=4.5=1_gnu
+  - _r-mutex=1.0.0=anacondar_1
+  - agate=1.7.1=py39h06a4308_0
```
pbiology, line 12: Shouldn't we remove the build specific information?

BALSAMIC/utils/cli.py
```diff
@@ -450,11 +450,19 @@ def bioinfo_tool_version_conda(
         conda_bioinfo_version: dict. A dictionary of tools as key and version as list in value
     """
     conda_bioinfo_version = current_bioinfo_tool_version
-    for p in packages:
-        if isinstance(p, dict):
+    for package in packages:
+        if isinstance(package, dict):
+            # Extraction of pip specific packages
+            bioinfo_tool_version_conda(
```
fevac, line 456: 🤯

fevac (commented):
Nice! 🌟  Did you figure out what was the issue before?

BALSAMIC/utils/cli.py
fevac: I guess this is done to retrieve this type of information?
```
 - pip:
      - attrs==22.2.0
      - click==6.7
      - configobj==5.0.8
```

BALSAMIC/containers/align_qc/align_qc.yaml
ivadym: I don't know... By specifying these platform-specific build restrictions we are fixing it to the containers image/platform. I assumed that it will be easier to resolve them this way, but I'm not sure. Maybe someone has a stronger opinion ❓

BALSAMIC/containers/vcf2cytosure/vcf2cytosure.yaml
ivadym: Great, thank you! ⚡ ⚡

BALSAMIC/utils/cli.py
ivadym: Yes. We were missing some of the tools like `cnvkit`, which will be now included in the config.json files (and the delivery report 🎈 )

BALSAMIC/containers/align_qc/align_qc.yaml
```diff
@@ -1,16 +1,194 @@
+name: base
```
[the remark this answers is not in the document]
ivadym: Sure, I will remove it :)

BALSAMIC/containers/align_qc/align_qc.yaml
fevac: I would remove the build specific tag (everything after the second = sign). I think keeping the version should be fixed enough and by removing the tag we avoid issues that could arise when building in a different platform (i.e. Mac). We could leave it without tags and revisit this in the future if we find we need to constrain it

BALSAMIC/containers/align_qc/align_qc.yaml
```diff
@@ -1,16 +1,194 @@
...
-  - bioconda::samtools=1.15.1
-  - bioconda::tabix=0.2.6
-  - bioconda::picard=2.27.1
-  - bioconda::multiqc=1.12
-  - bioconda::fastp=0.23.2
-  - conda-forge::csvkit=1.0.7
-  - conda-forge::libiconv=1.16
-  - conda-forge::fontconfig=2.14.0
-  - conda-forge::r-base=4.1.3
+  - _libgcc_mutex=0.1=main
```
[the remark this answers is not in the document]
ivadym: That's the thing, the platforms are fixed for us since we have it inside a container so it doesn't matter at the end. I'll try to find more about it, if not I will remove them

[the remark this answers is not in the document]
fevac: oh, I see your point. I guess it should not matter then. I guess this only would be an issue when creating the conda environment outside the container, which is not the intended use. Then I don't have strong opinions

ivadym:
> Nice! 🌟 Did you figure out what was the issue before?

This part was not flexible enough to allow the new dependency yaml structure:
https://github.com/Clinical-Genomics/BALSAMIC/blob/36a66e8a2b18f8aa687685a49dc65886840dc1b3/BALSAMIC/utils/cli.py#L453-L457

fevac (approved):
🚀

Issue #1123, opened 2023-03-29
SV variants "too long to annotate" from VEP are filtered out


mathiasbio:
**Describe the bug**

Looking at WGS tumor / normal case "notedshark", SV calls are disappearing during the "vep_somatic_sv" rule. 
The number of variants in the vcf-files before and after is different:
`vcf/SV.somatic.notedshark.svdb.research.vcf.gz` variants: 16152
`vep/SV.somatic.notedshark.svdb.research.vcf.gz` variants: 15844

This is a difference of 308 variants, which exactly corresponds to the number of warnings in the vep_somatic_sv stderr output from balsamic.
`grep WARNING BALSAMIC.notedshark.vep_somatic_sv.5.sh_4554097.err | wc -l`
= 308

And looking for an individual variant from the list of warnings, such as, "MantaDUP:TANDEM:335100:0:2:0:0:0:manta|DUP00079774:dellysv" you can find in the VCF before but not after this vep_somatic_sv rule.

I can only conclude that these variants when having this warning in VEP are in essence filtered out. This should not be the intended behaviour, and seems very urgent to fix! 

**To Reproduce**
Steps to reproduce the behavior.

**Expected behavior**

**If workflow, which rules**
If possible, and using the Snakemake workflows, the name of the affected rules and workflows.

**Screenshots**

<img width="1099" alt="image" src="https://user-images.githubusercontent.com/14308912/228481566-721b05bd-24d6-4e0d-a356-38e6a7aaf9c3.png">

**Version (please complete the following information):**
`balsamic --version` 11.2.0

**Additional context**
Perhaps this fix can be included in the new release of balsamic! 
https://github.com/orgs/Clinical-Genomics/projects/46
Perhaps added as an extra PR in this issue: https://github.com/Clinical-Genomics/BALSAMIC/issues/1119

mathiasbio:
It seems that this issue has been discussed on 'https://github.com/Ensembl/ensembl-vep/issues/600' and been fixed in the release >109 to still output the variants that are too long to annotate, instead of just filtering them away. 

Perhaps hen that a simple first step is just to update VEP and this issue could be solved, and hopefully the update doesn't cause any other issues.

fevac:
+1 VEP update

mathiasbio:
[...]

PR #1124, opened 2023-03-29
fix: VEP filters away variants too long to annotate

mathiasbio:
### This PR:

Aims to solve the issue mentioned here: https://github.com/Clinical-Genomics/BALSAMIC/issues/1123 where SV calls longer than the default max_sv_size are filtered out rather than simple being left un-annotated in the output VCF (in versions  > 109 of VEP this was changed so that variants are not filtered out).

Another way to keep all variants, as well as to annotate them, is simply to increase the max_sv_size to the maximum possible size (the size of chromosome 1: 249250621 for hg19, which is larger than chromosome 1 in hg38 and should therefore be fine for both genome versions).

**Changed: for changes in existing functionality.**
- [x] Changed --max_sv_size in VEP params to the size of chr1 for hg19. 

### Tests to be done

- [x] Install branch on Hasta
- [x] Run analysis on WGS T / N case hardyweevil
- [x] Analysis succeeding without errors
- [x] Variants filtered out from VEP with warning "too long to annotate" are kept in downstream VCF and annotated in case hardyweevil:
    - [x] 24 variants from vep_germline_normal are annotated, not filtered out
    - [x] 42 variants from vep_germline_tumor are annotated, not filtered out
    - [x] 224 variants from vep_somatic_sv are annotated, not filtered out
    - [x] remaining VEP annotations are unchanged

### Review and tests: 
- [x] Tests pass
- [x] Code review

### DIFF
--- a/BALSAMIC/constants/workflow_params.py
+++ b/BALSAMIC/constants/workflow_params.py
@@ -153,7 +153,7 @@ WORKFLOW_PARAMS = {
         "column_info": "-c 1 -S 2 -E 3 -g 4",
     },
     "vep": {
-        "vep_filters": "--compress_output bgzip --vcf --everything --hgvsg --allow_non_variant --dont_skip --buffer_size 30000 --format vcf --offline --variant_class --merged --cache --verbose --force_overwrite"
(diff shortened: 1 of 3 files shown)


ivadym (approved):
🥇

PR #987, opened 2022-08-26
feat: update TIDDIT SVDB delly vcf2cytosure and cnvkit containers

khurrammaqbool:
### This PR:

Changed: Update `TIDDIT` to v3.3.0, `SVDB` to v2.6.4, `delly` to v1.1.3, `vcf2cytosure` to v0.8.

### Review and tests: 
- [ ] Tests pass
- [ ] Code review
- [ ] New code is executed and covered by tests, and test approve

PR #1167, opened 2023-06-01
fix: delly version

ivadym:
### This PR:

The logged version of delly in the delivery report is incorrect (`v1.1.3`). We are running `delly v1.0.3.` 

Screenshot 2023-06-01 at 16 13 15

Delly version was meant to be updated in this PR. However, only the yaml file was changed:
- https://github.com/Clinical-Genomics/BALSAMIC/pull/987

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


mathiasbio:
If we add this to the 12.0.1 release patch, we might want to add some small sanity-check on SVs to make sure that Delly is still functioning properly 🤔

mathiasbio (approved):
Very nice catch! 🏎️

ivadym:
> If we add this to the 12.0.1 release patch, we might want to add some small sanity-check on SVs to make sure that Delly is still functioning properly 🤔

Since the results are likely to change, we are probably talking about a major release here and therefore we should merge it to develop and included in 13...

But I could definitely include the correct version in the 12.0.1, i.e. fix back the yaml file to `v1.0.3` 🤔

ivadym:
Closing. @khurrammaqbool reverted the version in #987 due to the significant changes introduced by that delly version. If we are determined to proceed with an update, our goal should be to implement the most recent release (currently v1.1.6).

ivadym:
The version logged to the delivery report was fixed here: https://github.com/Clinical-Genomics/BALSAMIC/pull/1170

PR #1117, opened 2023-03-22
refactor: reference workflows

ivadym:
### Changed:
- Reference workflows (hg19, hg38, canfam3, containers download)
- Refactoring of constants, models, and folder structure

### Review and tests: 
- [ ] Tests pass
- [ ] Code review
- [ ] New code is executed and covered by tests, and test approve

PR #851, opened 2022-01-24
feat: Option to use PON with TGA cnvkit

ashwini06:
### This PR:

Added: 
- Option to use PON reference (.cnn) if it is pre-existed while running TGA CNVkit **tumor-only** analysis. In the absence of PON reference file, the CNVkit workflow generates a Flatreference from genome files and uses it as a background

Removed: 
- Currently, sentieon TNhaplotyper tumor-only contains the option for using PON (.vcf), [this ](https://github.com/Clinical-Genomics/BALSAMIC/blob/f092cfe94005c0819108a2af5278fb8[…]ALSAMIC/snakemake_rules/variant_calling/sentieon_t_varcall.rule)rule accept pon file from config, but cli changes for config that it could accept pon vcf file was never possible. Alternative option is to supply with `--snakemake-opt` to add extra config keys. Anyways compared to existing downstream filters, PoN isn’t improving performance for SNV calling. So decision to remove this option - @hassanfa @ashwini06 

### Review and tests: 
- [x] Tests pass
- [ ] Code review
- [ ] New code is executed and covered by tests, and test approve

PR #1174, opened 2023-06-09
refactor: Constants

ivadym:
### This PR:

Part of https://github.com/Clinical-Genomics/BALSAMIC/pull/1117, which will be split in multiple and simpler PRs.

### Changed:
- Refactor constants

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


BALSAMIC/snakemake_rules/annotation/varcaller_wgs_filter_tumor_only.rule
```diff
@@ -41,7 +41,7 @@ bcftools +counts {output.vcf_pass_tnscope} > {output.bcftools_counts_research};
 rule bcftools_filter_tnscope_clinical_tumor_only:
   input:
     vcf_snv_clinical = vep_dir + "{var_type}.somatic.{case_name}.tnscope.clinical.vcf.gz",
-    wgs_calling_file = config["reference"]["wgs_calling_interval"]
+    wgs_calling_file = config["reference"]["wgs_calling_regions"]
```
mathiasbio, line 44: A small part of me finds it a bit misleading that the name is "wgs_calling" when it's filtering of already made calls. But the name from the source is "wgs_calling_regions" so I like that we're consistent in this too. In the end I think I approve of this change! 👍

BALSAMIC/snakemake_rules/quality_control/GATK.rule
```diff
@@ -6,7 +6,7 @@
 rule PreparePopVCF:
     input:
         bam = bam_dir + "tumor.merged.bam",
-        ref1kg = config["reference"]["1kg_snps_all"],
+        ref1kg = config["reference"]["vcf_1kg"],
```
mathiasbio, line 9: I like this renaming as the population vcf does also contain InDels, not only SNPs. 🥇

mathiasbio (approved):
Looks cleaner to me than the previous structure! Nice job!
☁️ ☁️ 🌕 ☁️ ☁️  
🔥 🌟 🧙 🌟 🔥 
🌲 🌲 🌲 🌲 🌲

khurrammaqbool (approved):
Thank you for taking up this huge refactoring. 
2. Generally we need to add some tests to cover the code. I have added a few suggestions. Please add them if convenient in this PR or open an issue so that we are aware of it and fix it at some point. e.g. if it is taking time to fix coverage for config/pon.py:L151-158 and config/case.py:L283-286, it would be nice to open an issue for it. The same applies to /utils/rule.py:L283-290. 
2. As I mentioned earlier, it would nice to make sure that the order of the constants is retained where relevant.

BALSAMIC/snakemake_rules/quality_control/GATK.rule
ivadym: Thanks for the review @mathiasbio <3

BALSAMIC/commands/run/analysis.py
```diff
@@ -228,7 +228,7 @@ def analysis(
         bind_path.append(sample_config.get("background_variants"))
     if "pon_cnn" in sample_config:
         bind_path.append(sample_config.get("panel").get("pon_cnn"))
```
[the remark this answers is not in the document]
ivadym: We've tried a couple of things already with Ashwini to cover that line (we have a bunch of commits testing different pytests here, including what you've just suggested: https://github.com/Clinical-Genomics/BALSAMIC/pull/851). But feel free to open a new PR if you feel confident that it will be solved 🙌

BALSAMIC/utils/class_types.py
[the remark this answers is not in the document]
ivadym: Actually it's not necessary because I end up testing it indirectly. If you have a look at the[ Balsamic init refactoring P](https://github.com/Clinical-Genomics/BALSAMIC/pull/1188)R, it's covered there. The problem that here I'm just defining the constants to later use them (drawbacks of splitting that huge PR): https://app.codecov.io/gh/Clinical-Genomics/BALSAMIC/pull/1188/blob/BALSAMIC/utils/class_types.py

BALSAMIC/commands/run/analysis.py
```diff
@@ -228,7 +228,7 @@ def analysis(
         bind_path.append(sample_config.get("background_variants"))
     if "pon_cnn" in sample_config:
```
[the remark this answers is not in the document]
ivadym: 👍  The whole analysis.py should be refactored. Will add it to the refactoring ideas issue

BALSAMIC/constants/cache.py
```diff
@@ -0,0 +1,454 @@
...
+    CNVPYTOR: str = "cnvpytor"
+    COVERAGE_QC: str = "coverage_qc"
+    DELLY: str = "delly"
+    PYTHON_3: str = "varcall_py3"
+    PYTHON_27: str = "varcall_py27"
+    SOMALIER: str = "somalier"
+    VCF2CYTOSURE: str = "vcf2cytosure"
+
+
+REFERENCE_FILES = {
```
fevac, line 83: I don't know about how I feel about this dictionary. Should it be a dictionary or a class?

And shouldn't we use the `ReferenceUrlsModel` within it?

fevac (commented):
Great work 🌟 !! Not many things from my side. Mostly the use of dictionaries to store info, but see the comments for details.

BALSAMIC/commands/run/analysis.py
```diff
@@ -230,7 +230,7 @@ def analysis(
...
         bind_path.append(sample_config.get("panel").get("pon_cnn"))
-    bind_path.append(BALSAMIC_SCRIPTS)
+    bind_path.append(SCRIPT_DIR)
     bind_path.append(sample_config["analysis"]["analysis_dir"])
```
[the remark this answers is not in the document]
ivadym: Yes, I totally agree, but this is a huge effort since it has to be aligned also with the `config case`, which is where these dictionaries are built.  Probably something that need to be address in it's own PR

BALSAMIC/constants/cache.py
ivadym: I think it should remain as a constant and therefore be a dictionary. Then in the Balsamic init PR we have a `ReferencesModels` where this dictionary is being fed for validation. However, it's a very good idea to make use of the model, thanks!; I will add the changes in the balsamic init PR since that method also has been refactored

ivadym: OK, we cannot use it in this way because there are circular imports involving the `ReferenceUrlsModel`. The cache models class relies on the constants that are defined in `constants/cache.py`. In the end, it actually makes sense not to initialise or validate the `REFERENCE_FILES` constant when defining it, but rather when we actually need to use it.

PR #1186, opened 2023-06-26
chore: Update add-to-bl action

pbiology:
### This PR:

Updates the github action which adds issues to the project. 
Just future proofing since I noticed it was a bit out of date.


pbiology:
Can we do this without bumping the version?

pbiology:
Actually it is not very urgent at all so perhaps it can just be part of next release

ivadym (approved):
Nice 👍  

We've been fixing deliberately the ubuntu version, but I suppose here it's fine to use `ubuntu-latest`

pbiology:
> Nice 👍
> 
> We've been fixing deliberately the ubuntu version, but I suppose here it's fine to use `ubuntu-latest`

Yeah I wasn't sure. I saw that in cg, servers etc ubuntu-latest was used. I guess to avoid an old version being deprecated. But I'm honestly not sure what is the best practice

ivadym:
Probably for the containers it's better to use a fixed version, but for the rest of the actions I think it's ok to use the latest

khurrammaqbool:
> > Nice 👍
> > We've been fixing deliberately the ubuntu version, but I suppose here it's fine to use `ubuntu-latest`
> 
> Yeah I wasn't sure. I saw that in cg, servers etc ubuntu-latest was used. I guess to avoid an old version being deprecated. But I'm honestly not sure what is the best practice

To my experience it is always nice to fix the OS version for docker container. For github actions, using latest is fine in case the action/repository is actively maintained. So it should be fine for most of the common repos to use `ubuntu-latest`.

ivadym:
Closing as the container version has already been updated in another PR

PR #1188, opened 2023-06-27
refactor: Balsamic init workflow

ivadym:
### This PR:

Part of https://github.com/Clinical-Genomics/BALSAMIC/pull/1117, which will be split in multiple and simpler PRs.

Changed:
- Balsamic init workflow refactoring 

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


BALSAMIC/models/cache.py
```diff
@@ -1,143 +1,453 @@
...
+            dbsnp=self.references.dbsnp.file_path + "." + FileType.GZ,
+            delly_exclusion=self.references.delly_exclusion.file_path,
+            delly_exclusion_converted=self.references.get_delly_exclusion_converted_file(),
+            delly_mappability=self.references.delly_mappability.file_path,
+            genome_chrom_size=self.references.genome_chrom_size.file_path,
+            gnomad_variant=self.references.gnomad_variant.file_path,
+            hc_vcf_1kg=self.references.hc_vcf_1kg.file_path + "." + FileType.GZ,
+            known_indel_1kg=self.references.known_indel_1kg.file_path
+            + "."
+            + FileType.GZ,
```
diitaz93, line 442: can't this be in the same line?

tests/conftest.py
```diff
@@ -7,13 +10,29 @@
...
+from BALSAMIC.constants.cache import (
+    DockerContainers,
+    GenomeVersion,
+    REFERENCE_FILES,
+)
+from BALSAMIC.constants.constants import FileType
 from _pytest.tmpdir import TempPathFactory
 
 from BALSAMIC.constants.cluster import ClusterConfigType
 from BALSAMIC.constants.paths import CONSTANTS_DIR
```
diitaz93, line 25: Would be nice to group all imports from balsamic

tests/conftest.py
```diff
@@ -22,6 +41,12 @@
 MOCKED_OS_ENVIRON = "os.environ"
 
 
+@pytest.fixture(scope="session", name="session_tmp_path")
+def fixture_session_tmp_path(tmp_path_factory: TempPathFactory) -> Path:
+    """Non existent file path."""
```
diitaz93, line 46:
```suggestion
    """Return a non-existent file path."""
```

tests/conftest.py
```diff
@@ -292,6 +319,30 @@ def swegen_sv_frequency(reference_variants_dir_path: str) -> str:
     return Path(reference_variants_dir_path, "swegen_sv.vcf.gz").as_posix()
 
 
+@pytest.fixture(scope="session", name="invalid_file")
+def fixture_invalid_file(session_tmp_path: Path) -> Path:
+    """Non existent file path."""
```
diitaz93, line 324:
```suggestion
@pytest.fixture(scope="session", name="invalid_json_file")
def fixture_invalid_json_file(session_tmp_path: Path) -> Path:
    """Return a non-existent json file path."""
```

tests/conftest.py
```diff
@@ -292,6 +319,30 @@ def swegen_sv_frequency(reference_variants_dir_path: str) -> str:
...
+    return Path(session_tmp_path, "invalid_file.json")
+
+
+@pytest.fixture(scope="session", name="json_file")
+def fixture_json_file(session_tmp_path: Path) -> Path:
+    """Mocked file path to use when writing JSON."""
```
diitaz93, line 330: For uniformity, I always start fixture docstrings with `Return`. But this is debatable

tests/conftest.py
```diff
@@ -918,3 +975,346 @@ def snakemake_fastqc_rule(tumor_only_config, helpers):
...
+@pytest.fixture(scope="session", name="cosmic_key")
+def fixture_cosmic_key() -> str:
+    """Mocked COSMIC key."""
+    return "ZW1haWxAZXhhbXBsZS5jb206bXljb3NtaWNwYXNzd29yZAo="
+
+
+@pytest.fixture(scope="session", name="cluster_account")
+def fixture_cluster_account() -> str:
+    """Mocked cluster account for job submission."""
+    return "development"
```
diitaz93, line 995: There is an ongoing debate about whether to use this kind of fixtures or just hard code them into the tests. If this is a constant defined somewhere else, it should be used instead (given that the context is the same). Otherwise, the string literal is preferred. This (I think) is to not overpopulate the contest files with simple fixtures.

tests/conftest.py
```diff
@@ -918,3 +975,346 @@ def snakemake_fastqc_rule(tumor_only_config, helpers):
...
+
+
+@pytest.fixture(scope="session", name="develop_containers")
+def fixture_develop_containers() -> Dict[str, str]:
+    """Develop containers fixture."""
```
diitaz93, line 1000: Or replace "dictionary" with something more meaningful, like mapping?
```suggestion
    """Return a Develop containers dictionary."""
```

BALSAMIC/workflows/reference.smk
```diff
@@ -1,422 +1,52 @@
 # syntax=python tabstop=4 expandtab
```
khurrammaqbool, line 1: Since we have reference_canfam3.smk, It may be nice to rename it as reference_hg.smk.

BALSAMIC/models/cache.py
```diff
@@ -1,143 +1,453 @@
...
+)
+from BALSAMIC.constants.constants import FileType, BwaIndexFileType
+from BALSAMIC.utils.exc import BalsamicError
 
 LOG = logging.getLogger(__name__)
 
 
-class ReferenceUrlsModel(BaseModel):
-    """Defines a basemodel for reference urls
+class AnalysisReferencesModel(BaseModel):
```
khurrammaqbool, line 18: It may be meaningful to rename it as ProcessReferencesModel as we are no analysing references. The same applies other reference models where applicable. Similarly, Reference or References should be either in the beginning, middle or at the end of names across all the models to make it coherent. The same applies to Hg and Canfam.

tests/conftest.py
```diff
@@ -918,3 +975,346 @@ def snakemake_fastqc_rule(tumor_only_config, helpers):
...
+    )
+    known_indel_1kg_file.touch()
+    return known_indel_1kg_file
+
+
+@pytest.fixture(scope="session", name="mills_1kg_file")
+def fixture_mills_1kg_file(session_tmp_path: Path) -> Path:
+    """Dummy Mills' high confidence InDels VCF file."""
+    mills_1kg_file: Path = Path(session_tmp_path, "variants", "mills_1kg_index.vcf.gz")
+    mills_1kg_file.touch()
```
diitaz93, line 1176: I am unsure about the clean-up of these files after the tests are done. I think using `tmp_path_factory` from `pytest` can ensure the proper clean-up of these files, but I am not very knowledgeable in this matter. Good to consider for a future PR

BALSAMIC/models/cache.py
```diff
@@ -1,143 +1,453 @@
...
-        """Validate file format according to constants"""
-        assert value in set(FileType), f"{value} not a valid reference file format."
-        return value
 
-    @validator("genome_version")
-    def check_genome_ver(cls, value) -> str:
-        """Validate genome version according constants"""
-        assert value in set(GenomeVersion), f"{value} not a valid genome version."
-        return value
+class ReferencesModel(BaseModel):
```
khurrammaqbool, line 111: Continuing from comment above, taking ReferenceModel as a starting point with naming, it may be coherent to add to it i.e. ReferenceModelUrl, ReferenceModelProcessHg, ReferenceModelProcessHg and so on.

khurrammaqbool (changes requested):
Great work refactoring the reference cache part of the workflow. I have some suggestions for naming conventions and I will take another round looking at the code later. 
You may choose to add it as references_* or *_references where applicable (see suggestions with the rule names). It may be better to name them by type with possibility to differentiate them e.g. if reference occurs as stand alone and along with other name then it will be better to add the type to reference_* as well. 
I saw the dag graph which looks nice, could you please also add some benchmarks after running this workflow locally as well as with SLURM options.
Regarding tmp directory with SLURM: I noticed that the tmp directory is very important. Please make sure the tmp directory is not /tmp or $HOME/tmp especially for singularity containers when using SLURM option.

diitaz93 (commented):
Great refactoring! 💯

ivadym:
> Great work refactoring the reference cache part of the workflow. I have some suggestions for naming conventions and I will take another round looking at the code later. You may choose to add it as references_* or __references where applicable (see suggestions with the rule names). It may be better to name them by type with possibility to differentiate them e.g. if reference occurs as stand alone and along with other name then it will be better to add the type to reference__ as well. I saw the dag graph which looks nice, could you please also add some benchmarks after running this workflow locally as well as with SLURM options. Regarding tmp directory with SLURM: I noticed that the tmp directory is very important. Please make sure the tmp directory is not /tmp or $HOME/tmp especially for singularity containers when using SLURM option.

Thanks for the review @khurrammaqbool! I'll add some benchmarks after running `balsamic init` again.

I think the naming convention you're proposing adds unnecessary verbosity (e.g. `/cache/vep_cache.rule`) since they already belong to uniquely identified folders (`cache` in this case).

Regarding the `/tmp` directory, for singularity it is only possibility in hasta. One can't pull containers from Dockerhub using `$HOME/tmp`: https://github.com/Clinical-Genomics/servers/pull/259

BALSAMIC/snakemake_rules/cache/reference_genome.rule
```diff
@@ -0,0 +1,72 @@
+"""Reference genome files processing rules."""
```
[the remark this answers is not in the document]
ivadym: 👍

BALSAMIC/snakemake_rules/cache/references.rule
```diff
@@ -0,0 +1,36 @@
+"""Common reference download rules."""
```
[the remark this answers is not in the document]
ivadym: 👍

BALSAMIC/workflows/reference.smk
ivadym: Sounds good! 👍

BALSAMIC/models/cache.py
ivadym: Maybe we don't need to include `Model` in the name at all. I had a look in CG and most of the pydantic models don't have `Model` in their name so I'll proceed to remove them 👍

BALSAMIC/models/cache.py
ivadym: Great suggestions! 👍 I have revised the names to maintain a consistent structure. However, it is generally not common practice to include verbs, such as "Process", in a class name. In this specific case, it would be inappropriate since the class does not actually process anything. Its main purpose is to serve as a representation of the `references.json` file.

BALSAMIC/commands/init/base.py
```diff
@@ -274,58 +158,52 @@ def initialize(
...
+    balsamic_run.account = account
+    balsamic_run.case_name = cache_config.analysis.case_id
     balsamic_run.cluster_config = cluster_config
-    balsamic_run.scheduler = get_schedulerpy()
+    balsamic_run.configfile = config_path.as_posix()
+    balsamic_run.forceall = force_all
+    balsamic_run.log_path = log_dir.as_posix()
+    balsamic_run.mail_type = mail_type
+    balsamic_run.mail_user = mail_user
     balsamic_run.profile = profile
```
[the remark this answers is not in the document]
ivadym: At first I added them but it seems that type hinting non-self attributes, outside the context of a class, is not directly supported:

> Non-self attribute could not be type hinted

BALSAMIC/commands/init/base.py
```diff
@@ -274,58 +158,52 @@ def initialize(
...
-    balsamic_run.forceall = force_all
-    balsamic_run.run_analysis = run_analysis
+    LOG.info("Starting reference generation workflow...")
+    balsamic_run: SnakeMake = SnakeMake()
+    balsamic_run.account = account
+    balsamic_run.case_name = cache_config.analysis.case_id
     balsamic_run.cluster_config = cluster_config
-    balsamic_run.scheduler = get_schedulerpy()
+    balsamic_run.configfile = config_path.as_posix()
+    balsamic_run.forceall = force_all
```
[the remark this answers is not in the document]
ivadym: Probably we should, but it's also a snakemake specific flag: `--forceall` 🤔 . Ignoring it for now

BALSAMIC/commands/init/base.py
```diff
@@ -274,58 +158,52 @@ def initialize(
...
-    balsamic_run.run_mode = run_mode
-    balsamic_run.forceall = force_all
-    balsamic_run.run_analysis = run_analysis
+    LOG.info("Starting reference generation workflow...")
+    balsamic_run: SnakeMake = SnakeMake()
+    balsamic_run.account = account
+    balsamic_run.case_name = cache_config.analysis.case_id
     balsamic_run.cluster_config = cluster_config
-    balsamic_run.scheduler = get_schedulerpy()
+    balsamic_run.configfile = config_path.as_posix()
```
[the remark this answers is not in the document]
ivadym: Same here: `--configfiles`

BALSAMIC/commands/init/base.py
```diff
@@ -274,58 +158,52 @@ def initialize(
...
-    balsamic_run = SnakeMake()
-    balsamic_run.working_dir = config_dict["output"]
-    balsamic_run.snakefile = snakefile
-    balsamic_run.configfile = config_json
-    balsamic_run.run_mode = run_mode
-    balsamic_run.forceall = force_all
-    balsamic_run.run_analysis = run_analysis
+    LOG.info("Starting reference generation workflow...")
+    balsamic_run: SnakeMake = SnakeMake()
+    balsamic_run.account = account
```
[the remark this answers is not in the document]
ivadym: That's a very good idea! Probably we would need to redefine the SnakeMake class, but I will add this as a point to the refactoring issue that I'm working on, thanks!

BALSAMIC/models/cache.py
```diff
@@ -1,143 +1,453 @@
...
+        refgene_bed       : RefSeq's gene BED format from UCSC
+        refgene_flat      : RefSeq's gene flat format from UCSC
+        refgene_txt       : RefSeq's gene txt format from UCSC
+    """
+
+    genome_chrom_size: FilePath
+    reference_genome: FilePath
+    refgene_bed: FilePath
+    refgene_flat: FilePath
+    refgene_txt: FilePath
```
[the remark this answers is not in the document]
ivadym: FilePath is a pydantic specific type that also performs validation. We want this model to fail if there is something that has not been downloaded 

> FilePath: like Path, but the path must exist and be a file

BALSAMIC/models/cache.py
```diff
@@ -1,143 +1,453 @@
...
+class AnalysisReferencesModel(BaseModel):
+    """
+    Reference files model for a general Balsamic analysis.
+
+    Attributes:
+        genome_chrom_size : genome chromosome sizes
+        reference_genome  : required field for reference genome FASTA file
+        refgene_bed       : RefSeq's gene BED format from UCSC
+        refgene_flat      : RefSeq's gene flat format from UCSC
+        refgene_txt       : RefSeq's gene txt format from UCSC
```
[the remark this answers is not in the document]
ivadym: Changed on all models! 💃

tests/conftest.py
ivadym: I agree 👍

BALSAMIC/models/cache.py
```diff
@@ -1,143 +1,453 @@
...
+
+    def get_gnomad_files(self) -> List[str]:
+        """Return gnomAD associated output files."""
+        return [
+            self.gnomad_variant.file_path,
+            self.gnomad_variant_index.file_path,
+        ]
+
+    def get_1k_genome_files(self) -> List[str]:
+        """Return 1000 Genome related files."""
```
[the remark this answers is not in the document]
ivadym: Good idea, I've changed it in all these methods 👍

BALSAMIC/models/cache.py
```diff
@@ -1,143 +1,453 @@
...
+        ]
+
+    def get_reference_by_path(self, reference_path: str) -> ReferenceUrlModel:
+        """Return a reference given its full path."""
+        for model in self.references:
+            reference: ReferenceUrlModel = model[1]
+            if reference.file_path == reference_path:
+                return reference
+        LOG.error(f"No reference with the provided reference path {reference_path}")
+        raise BalsamicError()
```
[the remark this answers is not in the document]
ivadym: We are specifying it it in the base command (the default level is `INFO`). But it's a good question, my experience is that whatever is passed to BalsamicError gets lost, and what is being printed is the `logging`. Same in CG, but maybe it has been fixed or changed recently 🤔

tests/conftest.py
ivadym: I agree, a constant is more appropriate here 👍

tests/conftest.py
ivadym: `session_tmp_path` is actually a `tmp_path_factory` directory 🥳

BALSAMIC/models/cache.py
ivadym: It has been formatted by black, but maybe I will make use of f strings. It will also look cleaner

fevac:
Sorry @ivadym I haven't managed to review this yet. I will prioritise it next week. Sorry about that

tests/conftest.py
ivadym: Move it to init yo mean as we do in CG? It's a very good idea 🙇‍♂️  I think I will do it in another PR as I want to further improve the constants we have in the code

BALSAMIC/commands/init/base.py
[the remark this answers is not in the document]
diitaz93: Is that just a warning? Maybe adding them won't add something to the compiler but it may make the file more readable

BALSAMIC/models/cache.py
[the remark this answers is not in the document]
diitaz93: Maybe they get caught?

tests/conftest.py
diitaz93: No, I meant putting all the imports from BALSAMIC one after the other in alphabetical order instead of having imports from other modules, like `click` or `.helpers`

BALSAMIC/commands/init/base.py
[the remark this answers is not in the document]
ivadym: I think the Snakemake has been incorrectly designed. We should be able to initialise the attributes when calling the class, but it's not possible so I believe that's what needs the refactoring

BALSAMIC/models/cache.py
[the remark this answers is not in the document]
ivadym: At least here it's not the case. The error is printed but without the provided message 🤷

tests/conftest.py
ivadym: Oh, I see! I just discover that pycharm has an automatic way to organise imports:

> `⌘Сmd` `⌥Opt` `⇧Shift`  `L`, select Optimize imports, and click Run.

fevac (commented):
I'm still half way with this review. But adding it so you can start having a look if you're in a rush. Very good job! nice improvement over existing code. I still have issues with the way balsamic is coded and how it doesn't take advantage of snakemake built-in features but I do realize that changing most of those things would require almost a full rewrite of balsamic. So I didn't comment on those.

BALSAMIC/commands/config/case.py
```diff
@@ -215,7 +215,7 @@ def case_config(
         balsamic_cache, balsamic_version, genome_version, "reference.json"
     )
     with open(reference_config, "r") as f:
```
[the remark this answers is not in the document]
ivadym: Good catch 👍

BALSAMIC/commands/init/base.py
```diff
@@ -1,270 +1,148 @@
...
-        docker_image_base_name = "release_v{}".format(container_version)
-    else:
-        docker_image_base_name = container_version
-
-    for image_suffix in set(DockerContainers):
-        container_stub_url = "{}:{}-{}".format(
-            DOCKER_URL, docker_image_base_name, image_suffix
+    if genome_version in [GenomeVersion.HG19, GenomeVersion.HG38] and not cosmic_key:
+        LOG.error(
+            f"No COSMIC authentication key specified. It is required when using {genome_version} reference."
```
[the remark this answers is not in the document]
ivadym: Let's not use them since we are also omitting them in CG

BALSAMIC/commands/init/base.py
```diff
@@ -1,270 +1,148 @@
...
-        )
+    if run_mode == RunMode.CLUSTER and not account:
+        LOG.error("A cluster account is required for cluster run mode")
         raise click.Abort()
 
-    if genome_version in ["hg38", "hg19"] and not cosmic_key:
-        LOG.error("cosmic db authentication key required with hg38 and hg19")
-        raise click.Abort()
-
-    # resolve outdir to absolute path
```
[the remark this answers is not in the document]
ivadym: CG will always provide an absolute path, but yes for manual submissions we have to make sure it's the absolute path we're using. Good call ⭐ !

BALSAMIC/constants/constants.py
```diff
@@ -0,0 +1,35 @@
...
+    GFF: str = "gff"
+    GTF: str = "gtf"
+    GZ: str = "gz"
+    JSON: str = "json"
+    LOG: str = "log"
+    SIF: str = "sif"
+    TBI: str = "tbi"
+    TSV: str = "tsv"
+    TXT: str = "txt"
+    VCF: str = "vcf"
```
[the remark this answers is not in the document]
ivadym: I would say it's better to have them that to not, although it could be a bit verbose 🤷

BALSAMIC/models/cache.py
```diff
@@ -1,143 +1,445 @@
...
+            self.reference_genome.file_path,
+            f"{self.reference_genome.file_path}.{FileType.FAI}",
+            self.reference_genome.file_path.replace(FileType.FASTA, FileType.DICT),
+        ] + self.get_reference_genome_bwa_index_file_paths()
+
+    def get_reference_genome_bwa_index_file_paths(self) -> List[str]:
+        """Return output BWA reference genome index files."""
+        bwa_index_files: List[str] = []
+        for bwa_index in BwaIndexFileType:
+            bwa_index_files.append(f"{self.reference_genome.file_path}.{bwa_index}")
```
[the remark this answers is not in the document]
ivadym: Yes! 💌

BALSAMIC/constants/rules.py
```diff
@@ -1,6 +1,19 @@
...
+common_cache_rules: List[str] = [
+    "snakemake_rules/cache/cadd.rule",
+    "snakemake_rules/cache/singularity_containers.rule",
+    "snakemake_rules/cache/delly.rule",
+    "snakemake_rules/cache/reference_genome_index.rule",
+    "snakemake_rules/cache/reference_download.rule",
+    "snakemake_rules/cache/refseq.rule",
+    "snakemake_rules/cache/reference_vcf.rule",
+    "snakemake_rules/cache/vep.rule",
+]
```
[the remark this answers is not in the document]
ivadym: you're right! 👍

BALSAMIC/commands/init/base.py
```diff
@@ -274,58 +152,52 @@ def initialize(
...
     try:
-        graph_pdf = graph_obj.render()
-        LOG.info("Reference workflow graph generated successfully - %s " % graph_pdf)
+        graph_pdf: Path = Path(graph.render())
+        LOG.info(
+            f"Reference workflow graph generated successfully ({graph_pdf.as_posix()}) "
+        )
     except Exception:
         LOG.error("Reference workflow graph generation failed")
         raise click.Abort()
```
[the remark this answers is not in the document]
ivadym: yes! In a future PR I'll also change it for the rest of the workflow

khurrammaqbool (approved):
I added only couple of additional suggestions as you already received comprehensive reviews. You may open issues to address them later in case it is out of scope for the current PR.

BALSAMIC/utils/cli.py
```diff
@@ -251,9 +242,9 @@ def get_snakefile(analysis_type, analysis_workflow="balsamic", reference_genome=
     snakefile = Path(p, "workflows", "balsamic.smk")
 
     if analysis_type == "generate_ref":
-        snakefile = Path(p, "workflows", "reference.smk")
+        snakefile = Path(p, "workflows", "reference_hg.smk")
         if "canfam3" in reference_genome:
-            snakefile = Path(p, "workflows", "reference-canfam3.smk")
+            snakefile = Path(p, "workflows", "reference_canfam3.smk")
```
fevac, line 247: I wonder why `canfam3` and not `canfam`. I see that in other places it's used `canfam` only. Any particular reason? If not, maybe using `canfam` across would be better

BALSAMIC/utils/cli.py
```diff
@@ -222,15 +222,6 @@ def createDir(path, interm_path=[]):
...
-def iterdict(dic):
-    """dictionary iteration - returns generator"""
-    for key, value in dic.items():
-        if isinstance(value, dict):
-            yield from iterdict(value)
-        else:
-            yield key, value
-
-
 def get_schedulerpy():
```
fevac, line 225:
```suggestion
def get_schedulerpy() -> str:
```

BALSAMIC/utils/cli.py
```diff
@@ -222,15 +222,6 @@ def createDir(path, interm_path=[]):
...
     """
     Returns a string path for scheduler.py
```
fevac, line 227:
```suggestion
    """Returns a string path for scheduler.py."""
```
Maybe this doc could be more descriptive. Maybe something like, "Returns the path where the slurm scheduler script is located." or similar

BALSAMIC/utils/io.py
```diff
@@ -1,15 +1,50 @@
...
+from graphviz import Source
+
+from BALSAMIC import __version__ as balsamic_version
+from BALSAMIC.utils.cli import CaptureStdout
+from BALSAMIC.utils.exc import BalsamicError
 
 LOG = logging.getLogger(__name__)
 
 
+def generate_workflow_graph(
```
fevac, line 18: 🤩

BALSAMIC/utils/io.py
```diff
@@ -35,3 +70,9 @@ def read_yaml(yaml_path: str) -> dict:
             return yaml.load(fn, Loader=yaml.SafeLoader)
     else:
         raise FileNotFoundError(f"The YAML file {yaml_path} was not found")
+
+
+def write_finish_file(file_path: str) -> None:
+    """Write finish file indicating the analysis completion."""
+    with open(file_path, mode="w") as finish_file:
+        finish_file.write("%s\n" % datetime.now().strftime("%Y-%m-%d %H:%M"))
```
fevac, line 78: f string?

BALSAMIC/workflows/PON.smk
```diff
@@ -61,11 +62,7 @@ rule all:
     output:
         pon_finish_file = pon_finish
     run:
-        import datetime
-
-        # PON finish timestamp file
-        with open(str(output.pon_finish_file), mode="w") as finish_file:
-            finish_file.write("%s\n" % datetime.datetime.now())
+        write_finish_file(file_path=output.pon_finish_file)
```
fevac, line 65: 🤩

BALSAMIC/workflows/reference_canfam3.smk
```diff
@@ -0,0 +1,48 @@
...
+
+LOG = logging.getLogger(__name__)
+
+# Balsamic cache configuration model
+cache_config: CacheConfig = CacheConfig.parse_obj(config)
+
+# Temporary directory and shell options
+os.environ["TMPDIR"] = cache_config.references_dir.as_posix()
+shell.executable("/bin/bash")
+shell.prefix("set -euo pipefail; ")
```
fevac, line 22: lines duplicated in `reference_hg.smk`, should we make a function to avoid duplication?

BALSAMIC/workflows/reference_canfam3.smk
fevac, line 1: I'm a bit confused, both `reference_canfam3.smk` and `reference_hg.smk` seem exactly the same to me with exception of the docstrings. Can we then have a general one?

tests/conftest.py
fevac, line 1: I only went quickly over the tests. INCREDIBLE AMOUNT OF WORK 💟 . So much better, readable and comprehensive 🤩 . Let me know if you want me go through them in detail to give you feedback otherwise I won't.

BALSAMIC/snakemake_rules/cache/reference_download.rule
```diff
@@ -0,0 +1,36 @@
+"""Common reference download rules."""
+
+
+rule download_references:
+    """Download Balsamic reference files."""
+    output:
+        reference_path="{reference_path}",
+    wildcard_constraints:
```
fevac, line 8: nice!

BALSAMIC/snakemake_rules/cache/reference_download.rule
fevac, line 1: So incredible nice that this is generalised and on a single place now. Amazing 🌟

BALSAMIC/snakemake_rules/cache/reference_download.rule
```diff
@@ -0,0 +1,36 @@
...
+        elif [[ "{params.reference.url}" == gs://* ]]; then
+            cmd="gsutil cp '{params.reference.url}' -"
+        else
+            cmd="wget '{params.reference.url}' -O -"
+        fi
+        if [[ "{params.reference.gzip}" == "True" ]]; then
+            cmd+=" | gunzip"
+        fi
+        eval "$cmd > '{output.reference_path}'" &> "{log}"
+        """
```
fevac, line 36: If I didn't missed anything we are removing the md5 checksum, is that right? Is that on purpose or unintentional?

BALSAMIC/snakemake_rules/cache/refseq.rule
```diff
@@ -0,0 +1,48 @@
...
+        """
+        # Generate RefSeq's BED file
+        (
+            header=$(awk -f "{params.refseq_script_path}" "{input.refgene_sql}")
+            (echo \"$header\"; cat "{input.refgene_txt}") |\
+                csvcut -t -c chrom,exonStarts,exonEnds,name,score,strand,exonCount,txStart,txEnd,name2 |\
+                csvformat -T |\
+                bedtools expand -c 2,3 |\
+                awk '$1~/chr[1-9]/ && $1!~/[_]/' |\
+                cut -c 4- |\
```
fevac, line 36: I'm a bit puzzle with this. As far as I see, both `refseq.rule` amd `refseq_canfam.rule` are the same except with the way they treat the chromosome naming.
- For humans, we remove the `chr` prefix when generating the bed file (with `cut -c 4-`) and we then modify the refseq files to also remove the `chr` prefix.
- In dogs we use the `chr` prefix. 

Could we unify this? Either use chr in both or removing in it both? that way we could also possibly unify these rules. BUT even if we cannot unify these rules, it would be nice to standarise the `chr` usage. Do you know the reason for this in the current version? Sorry if this has been mentioned somewhere else.

BALSAMIC/snakemake_rules/cache/reference_vcf.rule
```diff
@@ -0,0 +1,49 @@
...
+        f"{{vcf}}.{FileType.GZ}.benchmark.{FileType.TSV}"
+    log:
+        f"{{vcf}}.{FileType.GZ}.{FileType.LOG}",
+    shell:
+        """
+        bgzip "{input.vcf}" &> "{log}"
+        """
+
+
+rule index_vcfs:
```
fevac, line 30: omg I LOVE that these rules are now separated and using their own container ❤️

BALSAMIC/snakemake_rules/cache/delly.rule
```diff
@@ -0,0 +1,20 @@
+"""Delly reference specific rules."""
+
+
+rule convert_delly_exclusion_file:
+    """Remove all occurrences of "chr" from the Delly telomeres and centromeres exclusion file."""
```
fevac, line 5: Another `chr` manipulation here

BALSAMIC/utils/cli.py
```diff
@@ -531,25 +531,6 @@ def job_id_dump_to_yaml(job_id_dump: Path, job_id_yaml: Path, case_name: str):
         yaml.dump({case_name: jobid_list}, jobid_out)
 
 
-def get_md5(filename):
```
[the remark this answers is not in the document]
ivadym: Yes, something for the future 🌃

BALSAMIC/utils/rule.py
```diff
@@ -19,17 +19,6 @@
 LOG = logging.getLogger(__name__)
 
 
-def get_chrom(panelfile):
```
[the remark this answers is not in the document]
ivadym: Oh that could be a good explanation. But since it's unlikely we'll provide an incorrect bed file I think for now we can ignore it. The bed files are tested when added to the capture_bed_repository and not sure how much we'll gain adding it for the community

Btw, these methods have been added in 2018 (Balsamic v1 🤯 😆  )

BALSAMIC/utils/cli.py
ivadym: I made a constant out of `get_schedulerpy` thanks! 🥳

BALSAMIC/utils/io.py
ivadym: I will also change it for the rest of the workflow (but in a future PR 🌉 )

BALSAMIC/workflows/reference_canfam3.smk
ivadym: Oh I went completely blind here, such a nice suggestion thanks!

BALSAMIC/snakemake_rules/cache/reference_download.rule
ivadym: It was deliberate. We were not doing it properly, as we were calculating the MD5 checksum but failing to compare it with the downloaded file.

Furthermore, I made the decision to eliminate the associated methods because we were using weak hashing algorithms (hashlib.md5), which are known to be security-sensitive (https://docs.python.org/3/library/hashlib.html)

ivadym: It would be nice to have this in place in the future though. I will open an issue with some refactoring ideas and include them there.

BALSAMIC/snakemake_rules/cache/delly.rule
ivadym: Here is to make it compliant with what delly accepts as exclusion file so I think it's fine

BALSAMIC/utils/cli.py
ivadym: I merged the reference files upon your suggestion 🥳

fevac (approved):
🚀

BALSAMIC/snakemake_rules/cache/refseq.rule
khurrammaqbool: I may be able to clarify this a bit. The addition of contig i.e. chromosomes in case of chromosome level assembled genomes were not standardised. Since human genome was the first with two different assemblies i.e. [hg (numbers and names with "chr") and GRCh (without "chr")](https://gatk.broadinstitute.org/hc/en-us/articles/360035890951-Human-genome-reference-builds-GRCh38-or-hg38-b37-hg19#:~:text=Primary%20assembly,)%20and%20Mitochondrial%20(%20chrM%20).), until hg19 and GRCh37; other mammalian and avian genomes (almost all) opt to include chr with the chromosome names.  There is only one assembly after these two with the two are merged in GRCh38 going forward. Any dependent bed file we have so far added is actually for GRCh37, but we are wrongly calling it hg19 because the chromosome names are from GRCh37. Canfam or more precisely Canfam assembly 3 that we have added to the workflow contains chr because all the gene models and annotations etc. subsequently follow these names accordingly. Hence it is important to keep all the same (the assembly names annotation resources etc.) and aligned for the downstream analysis to work. I had proposed at one point to fix the name of the human assembly as well, but you may choose to rename it to GRCh37 were applicable or leave it as it is for now.

PR #1191, opened 2023-07-03
fix: pydantic version to <2.0

ivadym:
### This PR:

Pydantic v2, which introduced breaking changes, has been released: https://github.com/pydantic/pydantic/releases/tag/v2.0,  causing pytest and model misbehaviour.

### Changed:
- Fix Pydantic version (<2.0)

### Review and tests: 
- [ ] Tests pass
- [ ] Code review

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ requirements = [
     "numpy>=1.21.6",
     "pandas>=1.1.5",
     "psutil>=5.7.0",
-    "pydantic>=1.9.0",
+    "pydantic<2.0",
     "pygments>=2.6.1",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -21,7 +21,7 @@
     "numpy>=1.21.6",
     "pandas>=1.1.5",
     "psutil>=5.7.0",
-    "pydantic>=1.9.0",
+    "pydantic>=1.9.0,<2.0",
```
pbiology, line 24: Could one specify it as `"pydantic=1.9.*"`?

PR #1194, opened 2023-07-10
feat: Annotate SVs with somatic SV observation frequencies

khurrammaqbool:
### This PR:

Added: The annotation of somatic SVs from somatic SV observations frequencies.

### Review and tests: 
- [x] Tests pass
- [x] Code review


khurrammaqbool:
Screenshot 2023-07-18 at 11 05 39

BALSAMIC/snakemake_rules/annotation/vep.rule
```diff
@@ -70,135 +71,150 @@ rule vep_somatic_clinical_snv:
...
+    message_text = "SV.somatic." + config["analysis"]["case_id"] + ".svdb.research.vep.vcf.gz",
+  threads:
+    get_threads(cluster_config, "svdb_annotate_somatic_research_sv")
+  message:
+    "Running vep annotation for structural and copy number variants on {params.message_text}"
+  shell:
+      """
+if [[ -f "{params.swegen_sv_frequency}" ]]; then
+  svdb --query --bnd_distance 5000 --overlap 0.80 \
+  --in_occ OCC --out_occ swegen_obs --in_frq FRQ --out_frq SWEGENAF \
```
ivadym, line 151: Shouldn't we define the constants used here in `workflow_params.py`?

BALSAMIC/snakemake_rules/annotation/vep.rule
```diff
@@ -70,135 +71,150 @@ rule vep_somatic_clinical_snv:
...
-    shell:
-        """
-if [[ -f "{params.swegen_sv_frequency}" ]]; then
-svdb --query --bnd_distance 5000 --overlap 0.80 \
---in_occ OCC --out_occ swegen_obs --in_frq FRQ --out_frq SWEGENAF \
---db {params.swegen_sv_frequency} --query_vcf {input.vcf_sv_research} > {params.tmp_vcf_clinical};
 
-bgzip -l 9 -c {params.tmp_vcf_clinical} > {output.vcf_sv_clinical};
 
+rule svdb_annotate_somatic_research_sv:
```
ivadym, line 129: 4 spaces indentation

BALSAMIC/snakemake_rules/annotation/vep.rule
```diff
@@ -70,135 +71,150 @@ rule vep_somatic_clinical_snv:
...
+  bgzip -l 9 -c {params.vcf_somatic_obs} > {output.vcf_sv_clinical};
 else
-
-cp {input.vcf_sv_research} {output.vcf_sv_clinical};
-
+  cp {input.vcf_sv_research} {output.vcf_sv_clinical};
 fi
 
 tabix -p vcf -f {output.vcf_sv_clinical};
         """
```
ivadym, line 208: Is it possible to simplify this part and reduce code duplication using individual ifs and overwriting the output file?
```
shell:
  """
if [[ -f "{params.clinical_sv_observations}" ]]; then
...

if [[ -f "{params.somatic_sv_observations}" ]]; then
...

  """
```

BALSAMIC/snakemake_rules/annotation/vep.rule
```diff
@@ -70,135 +71,150 @@ rule vep_somatic_clinical_snv:
...
 --dir_plugins $vep_path \
---output_file {output.vcf_research} \
+--output_file {output.vcf_research_vep} \
 --fork {threads} \
 {params.vep_defaults} \
 
-tabix -p vcf -f {output.vcf_research};
-    """
+tabix -p vcf -f {output.vcf_research_vep};
+          """
```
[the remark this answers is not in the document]
ivadym: Just review the indentation of the rules you've added 👍

BALSAMIC/snakemake_rules/annotation/vep.rule
```diff
@@ -70,135 +71,150 @@ rule vep_somatic_clinical_snv:
...
 tabix -p vcf -f {output.vcf_snv_clinical};
     """
 
-rule vep_somatic_sv:
+
+rule vep_somatic_research_sv:
   input:
-    vcf_research = vcf_dir + "SV.somatic.{case_name}.svdb.research.vcf.gz",
-    header = vcf_dir + "SV.somatic.{case_name}.svdb.sample_name_map",
+    vcf_research = vcf_dir + "SV.somatic." + config["analysis"]["case_id"] + ".svdb.research.vcf.gz",
```
ivadym, line 94: I think we should start using f strings from now on for inputs/outputs, messages, etc. Makes the code more compact and easier to follow

ivadym (commented):
Nice, looks good 👍

BALSAMIC/snakemake_rules/annotation/vep.rule
khurrammaqbool: The solution you propose can not be implemented until we perform multiple operation on the same input file. I suggest we keep current solution and work on refactoring in separate PR if a better solution is found.

BALSAMIC/snakemake_rules/annotation/vep.rule
khurrammaqbool: Will be addressed in https://github.com/Clinical-Genomics/BALSAMIC/issues/1208

BALSAMIC/snakemake_rules/annotation/vep.rule
khurrammaqbool: This will be addressed in https://github.com/Clinical-Genomics/BALSAMIC/issues/1209

khurrammaqbool:
> Nice, looks good 👍

Thank you @ivadym for the review. I have addressed almost all of them and opened issues that fell outside the scope of this PR. Although some of the refactoring could also be taken to another PR, but I made those changes here.

BALSAMIC/snakemake_rules/annotation/vep.rule
ivadym: 👍

BALSAMIC/snakemake_rules/annotation/vep.rule
ivadym: 👍

BALSAMIC/snakemake_rules/annotation/vep.rule
ivadym: I would rather do that than having code duplication. It also makes the code more complex and difficult to understand and we are, in any case, marking the intermediate files as temporary

khurrammaqbool: I split this in two separate rules, instead. This will avoid performing multiple operations on same input file, which was not possible to implement with the proposed alternative.

ivadym: Oh this is even more elegant, I LOVE IT 💯

BALSAMIC/constants/cluster_analysis.json
```diff
@@ -192,10 +192,22 @@
...
+		"time": "12:00:00",
+		"n": 8
+	},
+	"svdb_annotate_clinical_obs_somatic_clinical_sv": {
+		"time": "10:00:00",
+		"n": 8
+	},
+	"svdb_annotate_somatic_obs_somatic_clinical_sv": {
+		"time": "10:00:00",
+		"n": 8
```
ivadym, line 209: This is complicated: I like that we are following our naming structure... but this has completely lost meaning for me (same as the Balsamic word 🤣 )

What do you think about something like this:
`svdb_annotate_clinical_sv_observations`
`svdb_annotate_somatic_sv_observations`

Doesn't seem that we are annotating research variants and if we do, probably it will be ok to just include research in the name.

tests/conftest.py
```diff
@@ -860,41 +910,53 @@ def sample_config(tumor_sample_name: str, normal_sample_name: str):
...
+    """
     return "tests/test_data/qc_files/analysis"
 
 
 @pytest.fixture(scope="session")
 def multiqc_data_path(analysis_path):
-    """multiqc_data.json test path"""
+    """
+    Return path of JSON for MultiQC test data
+    """
```
ivadym, line 923: The idea is to have one-line docstrings where they fit. We need to refactor the existing ones, not the new ones that have the correct format.

```suggestion
    """Return path of JSON for MultiQC test data."""
```

BALSAMIC/snakemake_rules/annotation/vep.rule
```diff
@@ -83,122 +82,158 @@ if [[ -f "{params.clinical_snv}" || -f "{params.cancer_germline_snv}" || -f "{pa
...
---in_occ Obs --out_occ clin_obs --in_frq Frq --out_frq Frq \
---db {params.clinical_sv_observations} --query_vcf {input.vcf_sv_research} > {params.tmp_vcf_clinical}
-
-bgzip -l 9 -c {params.tmp_vcf_clinical} > {output.vcf_sv_clinical};
-
+if [[ -f "{params.somatic_sv_observations}" ]]; then
+    svdb --query --bnd_distance 10000 --overlap 0.80 \
+    --in_occ Obs --out_occ somatic_obs --in_frq Frq --out_frq somatic_frq \
+    --db {params.somatic_sv_observations} --query_vcf {input.vcf_sv_clinical_obs} > {params.vcf_somatic_obs}
+    bgzip -l 9 -c {params.vcf_somatic_obs} > {output.vcf_sv_clinical};
```
ivadym, line 218: ⭐

ivadym (approved):
Nice 💯 

Thanks for addressing my comments @khurrammaqbool  🥇

BALSAMIC/constants/cluster_analysis.json
khurrammaqbool: I agree with you @ivadym. I tried my best to adhere to the the [format defined for rule names](https://balsamic.readthedocs.io/en/latest/snakemake_etiquette.html). It is always good to always follow what is define and at least be consistent in case we have not already defined something. This will ensure understanding the overall structure, workflow as well as the code. If we need to deviate because of some reason then it will be nice to revise that in etiquette so that it is applicable to all for the reason mentioned above. I suggest we modify the rule name format: 
from `<program>_<function>_<optional_tag>_<optional_tag>` 
to `<program>_<bioinformatics_method>_<effected_output>` 
replacing  program with the bioinformatics method/process and any optional tags with the outcome or effect of the process in terms of the output highlighting the modified behaviour. This will make the names look like this:
`svdb_annotate_sv_clinical_observations`
`svdb_annotate_sv_somatic_observations`
I will open another issue to update the docs and update the rule names accordingly if you agree.

tests/conftest.py
khurrammaqbool: Continuing from the reply about the consistency above, would it not be nice to keep the format consistent across the code, which will make the format easy to follow?

ivadym: Yes, and the way to do it is following the coding guidelines: https://peps.python.org/pep-0257/#one-line-docstrings

BALSAMIC/constants/cluster_analysis.json
ivadym: I think what I proposed kind of follows that structure, but because it too general. So yes, feel free to open an issue.  I like your suggestion and It's always a good idea to have it as standardise as possible

khurrammaqbool: Will be addressed in https://github.com/Clinical-Genomics/BALSAMIC/issues/1214

PR #1206, opened 2023-07-19
fix: Balsamic container & pytests (fix cyvcf2 version)

ivadym:
### This PR:

The recent Cython update to v0.30 has led to compilation errors, resulting in the failure to execute the Balsamic container and, more importantly, preventing pytest from running via GitHub actions.

Screenshot 2023-07-19 at 14 51 37

### Changed:
- Fix cyvcf2 to version 0.30.22

### Review and tests: 
- [x] Tests pass
  Screenshot 2023-07-20 at 13 40 04
- [ ] Code review

### DIFF
--- a/.github/workflows/docker_build_push.yml
+++ b/.github/workflows/docker_build_push.yml
@@ -1,9 +1,9 @@
 name: Build and push master and develop container
 on:
   push:
-    branches:
-      - master
-      - develop
+#    branches:
+#      - master
+#      - develop
 jobs:
   main:
     name: Docker image build push
--- a/.github/workflows/docker_build_push.yml
+++ b/.github/workflows/docker_build_push.yml
@@ -1,9 +1,9 @@
 name: Build and push master and develop container
 on:
   push:
-#    branches:
-#      - master
-#      - develop
+    branches:
+      - master
+      - develop
 jobs:
   main:
     name: Docker image build push
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,8 +1,8 @@
 [12.0.2]
 --------
 
-Fixed:
-^^^^^^
+Changed:
+^^^^^^^^
 * Fix cyvcf2 to version 0.30.22 https://github.com/Clinical-Genomics/BALSAMIC/pull/1206
 
 [12.0.1]
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,10 @@
+[12.0.2]
+--------
+
+Fixed:
+^^^^^^
+* Fix cyvcf2 to version 0.30.22 https://github.com/Clinical-Genomics/BALSAMIC/pull/1206
+
 [12.0.1]
 --------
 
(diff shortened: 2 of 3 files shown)


mathiasbio (commented):
Sorry for the slow review! I have no comments about the code.

For the cyvcf2 changes, I'm not sure if I understand the changes correctly. If there has been no change to the version you can ignore my comments below. 

If the version has been changed I'm wondering if any tests are required to make sure that the following scripts are working properly, and do not affect the results in any negative unforeseen way.

https://github.com/Clinical-Genomics/BALSAMIC/blob/2c6a33e9e5478ff72c76cd718fc6c792f34426bb/BALSAMIC/commands/plugins/vcfutils.py#L9

https://github.com/Clinical-Genomics/BALSAMIC/blob/2c6a33e9e5478ff72c76cd718fc6c792f34426bb/BALSAMIC/assets/scripts/edit_vcf_info.py#L8

It seems for instance that edit_vcf_info.py is used in varcaller_filter_tumor_only.rule, and in varcaller_filter_tumor_normal.rule

And it could be nice to confirm that the VCF-files are not affected in any major way, just to avoid the possibility of needing to validate for TGA too in this small patch, since these rules affect the TGA workflow. 

For the vcfutil plugin I don't know where it's used exactly! But similarly there, have we considered effects on this plugin based on the changes?

mathiasbio (approved):
Meant to approve along with the comments.
Nice fix as always! The hero we don't deserve but are happy we have regardless ⭐ 🧙 ⭐

ivadym:
> Sorry for the slow review! I have no comments about the code.
> 
> For the cyvcf2 changes, I'm not sure if I understand the changes correctly. If there has been no change to the version you can ignore my comments below.
> 
> If the version has been changed I'm wondering if any tests are required to make sure that the following scripts are working properly, and do not affect the results in any negative unforeseen way.
> 
> https://github.com/Clinical-Genomics/BALSAMIC/blob/2c6a33e9e5478ff72c76cd718fc6c792f34426bb/BALSAMIC/commands/plugins/vcfutils.py#L9
> 
> https://github.com/Clinical-Genomics/BALSAMIC/blob/2c6a33e9e5478ff72c76cd718fc6c792f34426bb/BALSAMIC/assets/scripts/edit_vcf_info.py#L8
> 
> It seems for instance that edit_vcf_info.py is used in varcaller_filter_tumor_only.rule, and in varcaller_filter_tumor_normal.rule
> 
> And it could be nice to confirm that the VCF-files are not affected in any major way, just to avoid the possibility of needing to validate for TGA too in this small patch, since these rules affect the TGA workflow.
> 
> For the vcfutil plugin I don't know where it's used exactly! But similarly there, have we considered effects on this plugin based on the changes?

Thanks for the review 💟 

We need to validate the entire workflow as we had to address issues with the balsamic container by fixing the pydantic version. I think it's fine with the integration tests.


Regarding `vcfutil.py`,  we are only using it for `createvcf()` method, which is actually not being used anywhere except for the pytests. 

But we're using it for [edit_vcf_info.py](https://github.com/Clinical-Genomics/BALSAMIC/blob/2c6a33e9e5478ff72c76cd718fc6c792f34426bb/BALSAMIC/assets/scripts/edit_vcf_info.py#L8) as you said, but I'm hoping that the pytest we have is reliable enough. Doen't seem we have breaking changes with the new version of cyvcf2 and I would say we validate it as part of the integration tests

PR #1216, opened 2023-08-04
perf: Update python version and its dependencies

ivadym:
## Changed:
- Updated Python version to 3.11 and its dependencies 

### Review and tests: 
- [x] Tests pass
  - [x] Balsamic container is working Screenshot 2023-08-04 at 15 11 37
  - [x] CNV report is correctly rendered (updated deprecated method)
  [new_report.pdf](https://github.com/Clinical-Genomics/BALSAMIC/files/12290689/new_report.pdf)
  [old_report.pdf](https://github.com/Clinical-Genomics/BALSAMIC/files/12290690/old_report.pdf)
  - [x] Integrity test:
    - [x] TN-WGS case (`fleetjay`)
    - [x] TO-panel case (`setamoeba`)

- [ ] Code review


ivadym:
Ready for review @Clinical-Genomics/cancer-bioinfo!

After I merge it into develop, please make sure you update your environment accordingly

setup.py
```diff
@@ -1,51 +1,133 @@
...
     requirements.extend(["cyvcf2==0.30.22"])
 
 setup(
-    name="BALSAMIC",
-    version=about["__version__"],
-    url="https://github.com/Clinical-Genomics/BALSAMIC",
-    author="Hassan Foroughi Asl",
-    author_email="hassan.foroughi@scilifelab.se",
+    name=NAME,
+    version="12.0.2",
```
fevac, line 125: do we want to have this fixed?

tests/commands/init/test_init.py
```diff
@@ -176,7 +177,7 @@ def test_init_hg_graph_exception(
     # GIVEN a temporary output directory and a COSMIC key
 
     # WHEN invoking the init command
-    with mock.patch.object(Source, "render"):
+    with mock.patch.object(Source, "render", side_effect=BalsamicError("Test error")):
```
fevac, line 180: I didn't know this

tests/commands/init/test_init.py
```diff
@@ -189,7 +190,7 @@ def test_init_hg_graph_exception(
             ]
         )
 
-    # THEN the human reference generation workflow should have successfully started
+    # THEN the human reference generation workflow should have been canceled
```
fevac, line 193: should fail?

fevac (changes requested):

fevac:
Nice! 🌟 I forgot to add the review comment at the end 😬 I'm skeptical about fixing the version string. The rest looks good to me!

.readthedocs.yml
```diff
@@ -7,7 +7,7 @@ version: 2
 build:
   os: "ubuntu-20.04"
   tools:
-    python: "3.7"
+    python: "3.11"
```
khurrammaqbool, line 10: I suggest fixing the minor version as well to to avoid fails/debugging after minor versions automatically updating during builds.

setup.py
```diff
@@ -1,51 +1,133 @@
...
+    "rsa==4.7.2",
+    "six==1.16.0",
+    "smart-open==6.3.0",
+    "smmap==5.0.0",
+    "snakemake==7.32.0",
+    "stopit==1.1.2",
+    "tabulate==0.9.0",
+    "throttler==1.2.2",
     "toml==0.10.2",
+    "tomli==2.0.1",
```
khurrammaqbool, line 106: Could you please run a case to make sure that the behaviour of toml functions is not changed.

khurrammaqbool (approved):
Nice work 👍 
I have some minor comments. I suggest fixing python down to the minor version and please make sure `tomli` behaves as intended with `toml` by running a case. Since python version effects many dependencies, it will be nice to test the installation of `BALSAMIC` in a new conda environment to make sure all the dependencies are resolved.

tests/commands/init/test_init.py
ivadym: 👍

setup.py
ivadym: We are not fixing it, if you check the updated `.bumpversion.cfg` file, I've added this file there. So now it will be automatically updated to the deployment version when making the release.

https://github.com/Clinical-Genomics/BALSAMIC/blob/fd98ed2bf7df84eda878a868a727cdf3c39dfb95/.bumpversion.cfg#L24-L28

ivadym: In CG I believe we follow the same strategy 😇

setup.py
ivadym: I've run a TN-WGS and a TN-panel case, and both finished successfully. The `toml` version is the same, I've just added more dependencies to the setup.py :)

.readthedocs.yml
ivadym: Then we'll have to require users to have a very strict version of python ("==3.11.0"), which I don't think it's a good idea. Probably a better solution is to keep our dependancies up to date and revisit them more often.

setup.py
fevac: the things I learn from you!! 🌟

fevac (approved):
🚀

.readthedocs.yml
khurrammaqbool: As you mentioned "revisit them", I am thinking the same. Fixing the "major.minor.patch" for python, and dependencies, with the current "tested" version when we revisit the dependencies. We should manually update this instead of letting it updated automatically by not fixing it down to patch version. As we have seen in the past that sometimes the patch version can fail and we have to wait at least a couple of days for the developers to fix it if it originates in those dependencies, which could happen while deploying as well. This will potentially cause unnecessary delays for development, testing and deployment and compromise the robustness to some extent. I agree that the downside will be to visit it again, but we were using 3.7 for very long time until it is deprecated so it may be nice to have some schedule for revisiting the versions instead of only fixing them when they require fixing or get deprecated.  
Also, we have fixed all the dependencies to patch level. The same argument applies to the version of python as well. There is a tradeoff between having robustness vs using the latest, which could be balanced with proper implementation of schedules for the updates.

ivadym: I believe its significance in this context is somewhat diminished, primarily because we have the Python version locked in our environment. For GitHub actions or docs, the only places where the Python version isn't fixed, I think it's unlikely that an updated version will break them. If you have a strong opinion regarding this we can change it

PR #1218, opened 2023-08-07
fix: SVDB annotation intermediate rule 

ivadym:
### This PR:

Screenshot 2023-08-07 at 08 59 03

## Fixed:
* Fix SVDB annotation intermediate rule 

### Review and tests: 
- [ ] Tests pass
- [ ] Code review

### DIFF
--- a/BALSAMIC/snakemake_rules/annotation/vep.rule
+++ b/BALSAMIC/snakemake_rules/annotation/vep.rule
@@ -223,7 +223,6 @@ fi
 tabix -p vcf -f {output.vcf_sv_clinical};
 
 rm {input.vcf_sv_clinical_obs};
-rm {params.vcf_somatic_obs};
         """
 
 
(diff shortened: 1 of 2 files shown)


BALSAMIC/snakemake_rules/annotation/vep.rule
```diff
@@ -223,7 +223,6 @@ fi
 tabix -p vcf -f {output.vcf_sv_clinical};
 
 rm {input.vcf_sv_clinical_obs};
-rm {params.vcf_somatic_obs};
```
khurrammaqbool, line 226: Please move it to before else (L219) and after L218.

ivadym: Since vcf_somatic_obs is tagged as temp(), wouldn't it be automatically removed once the rule is completed?

`vcf_somatic_obs = temp(vep_dir + "SV.somatic." + config["analysis"]["case_id"] + ".svdb.somatic_obs.vcf"),`

khurrammaqbool: It did not remove in my tests that is why I added this line, only that I should have added it within the condition.

ivadym: You're right, we are not deleting temp files actually:

> --delete-temp-output
> Remove all temporary files generated by the workflow. Use together with –dry-run to list files without actually deleting anything. Note that this will not recurse into subworkflows.
> 
> Default: False

I will go with your solution for now then and create an issue to address it, thanks!

ivadym: Done @khurrammaqbool 👍

khurrammaqbool (approved):

BALSAMIC/snakemake_rules/annotation/vep.rule
khurrammaqbool: We retain many intermediate files after the completion of workflow.

PR #1071, opened 2023-01-24
fix: Build and push containers via GitHub actions

ivadym:
### This PR:

The containers are pushed to docker hub as a manifest index, not as an actual image. This seems to be due to a recent release of [buildx](https://github.com/docker/buildx/releases/tag/v0.10.0).

> Buildx v0.10 enables support for a minimal [SLSA Provenance](https://slsa.dev/provenance/) attestation, which requires support for [OCI-compliant](https://github.com/opencontainers/image-spec) multi-platform images. This may introduce issues with registry and runtime support (e.g. https://github.com/docker/buildx/issues/1533). You can optionally disable the default provenance attestation functionality using --provenance=false.

Proposed solutions:
- [x] Setting `provenance: false` ✅ 

or

- [x] Reverting to` v0.9.1`  ✅ 

### Fixed:
- Container build and push via github actions

### Review and tests: 
- [ ] Tests pass
- [ ] Code review
- [ ] New code is executed and covered by tests, and test approve

PR #1221, opened 2023-08-09
refactor: Update Docker containers Github Actions

ivadym:
### This PR:

Updates how and when we build and publish docker containers.

Proposed structure:
- Removed build and publish of containers in `master` branch.
- `develop` containers are build, published, and tested when:
  -  Pushed to `develop` branch.
  - In `pull_requests`, if any changes are made to the containers. They will be published with the `develop` tag (we won't need to build them locally anymore for testing purposes 🥳 ).
- `release` container will be build and published when the working branch matches the `release_v*` branch. This will ensure that we are performing verifications, validations, and deployment using the same containers.

> [!IMPORTANT] 
> Documentation has also been updated accordingly.

### Review and tests: 
- [x] Tests pass
  - [x] No containers are build or published by default [cf2e645](https://github.com/Clinical-Genomics/BALSAMIC/pull/1221/commits/cf2e645c5baedb23200c61c0fe1acc717d5636e3)
  - [x] No containers are build or published when base branch is master and the branch does not follow release convention [49620d6](https://github.com/Clinical-Genomics/BALSAMIC/pull/1221/commits/49620d69537f5d82e6fe0d7a5958f3895d99b4a4)
  - [x] Develop container build & publish triggered after changes made to `containers` directory [1d4a15a](https://github.com/Clinical-Genomics/BALSAMIC/pull/1221/commits/1d4a15a1cfe599e6eb563715bd7b0e39b8b71a32)
  - [x] Release container build & publish triggered by matching branch name [906be65](https://github.com/Clinical-Genomics/BALSAMIC/pull/1221/commits/906be6512e9e79aae7696924e19aa0bc8c2c3c96)
- [ ] Code review


.github/workflows/docker_build_publish.yml
```diff
@@ -1,9 +1,13 @@
...
   push:
     branches:
-      - master
-      - develop
+      - "develop"
+  pull_request:
+    paths:
+      - "BALSAMIC/containers/**"
+    branches-ignore:
+      - "release_*"
```
fevac, line 10:
```suggestion
    branches-ignore:
      - "release_*"
      - "master"
```
do we need to ignore `master` too?

.github/workflows/docker_build_publish.yml
```diff
@@ -36,6 +36,14 @@ jobs:
...
           build-args: CONTAINER_NAME=${{ matrix.container-name }}
           provenance: false
+      - name: Test container
+        id: docker_test_container
+        shell: bash
+        run: |
+          docker run \
+          -v $(readlink -f container_tests):/container_tests \
+          clinicalgenomics/balsamic:develop-${{ matrix.container-name }} \
+          bash /container_tests/${{ matrix.container-name }}/${{ matrix.container-name }}.sh
```
fevac, line 49: Sorry, a bit confused with this, what test does this run?

fevac: is this to avoid having corrupted containers being uploaded without noticing?

.github/workflows/docker_build_publish.yml
```diff
@@ -1,9 +1,13 @@
...
-      - develop
+      - "develop"
+  pull_request:
+    paths:
+      - "BALSAMIC/containers/**"
+    branches-ignore:
+      - "release_*"
 jobs:
   main:
     name: Docker image build push
```
fevac, line 13:
```suggestion
    name: Docker image build and publish to develop
```
Do we want to be more specific here?

.github/workflows/docker_build_publish_release.yml
```diff
@@ -1,18 +1,14 @@
...
-      - develop
-    tags-ignore:
-      - "**"
+  pull_request:
+    branches:
+      - "release_*"
 jobs:
   main:
-    name: Docker image build test
+    name: Docker image build push
```
fevac, line 8:
```suggestion
    name: Docker image build and publish
```

.github/workflows/docker_build_publish_release.yml
```diff
@@ -22,21 +18,26 @@ jobs:
...
+      - name: Build and push
+        id: docker_build_push
         uses: docker/build-push-action@v3
         with:
           file: BALSAMIC/containers/${{ matrix.container-name }}/Dockerfile
-          push: false
+          push: true
           tags: clinicalgenomics/balsamic:${{ steps.get_branch_name.outputs.branch }}-${{ matrix.container-name }}
           build-args: CONTAINER_NAME=${{ matrix.container-name }}
+          provenance: false
```
fevac, line 40: why would we want this set as false? don't we want to keep metadata for traceability?

docs/FAQs.rst
```diff
@@ -66,33 +66,25 @@ Our release model looks like in the figure below:
 
 .. figure:: images/git_releasemodel.png
 
-**Requirements:** bumpversion
+If changes are hotfixes/patches, make a release from the `master` branch. Otherwise, make a release from the `develop`
```
fevac, line 69: So basically, regardless of its a proper release or a hotfix, we will need to rebuild containers and cache, right?

docs/FAQs.rst
```diff
@@ -66,33 +66,25 @@ Our release model looks like in the figure below:
...
-Make a pull request to master at this point. After pull request is approved and merge it into master:
-
-1. `git checkout master`
-2. `git pull`
-3. `bumpversion --verbose [major/minor/patch]`
-4. `git push`
-5. `git push --tags`
+#. Switch to the `develop` branch: `git checkout develop`.
+#. Fetch the latest changes from `master` into `develop`: `git pull origin master`.
+#. Resolve any conflicts, commit changes, and push back to `develop`.
```
fevac, line 76: I did find clearer the previous way for these points:

```
1. `git checkout master && git pull`
2. `git checkout develop`
3. `git merge master`
4.  Fix all possible conflicts and `git push`
```

.github/workflows/docker_build_publish_release.yml
fevac, line 1: should this file be renamed to `docker_build_publish_release_develop.yml`

fevac (commented):
Awesome! loving where balsamic is going ❤️

.github/workflows/docker_build_publish.yml
ivadym: Mmm good question... Could there be a situation in which the base branch for pull requests is `master`? I can't think of such a scenario, anything I'm missing?

fevac: That being said I don't think they should be modifying containers but as a sanity check?

.github/workflows/docker_build_publish.yml
ivadym: yes! we have a folder named `container_tests`, with several bash scripts that we run through this GitHub action. it's not so efficient, since often we can invoke the tools (this test pass) but sometimes they provide error logs that go unnoticed.

.github/workflows/docker_build_publish.yml
ivadym: But if we have hotfixes, the PR will have `release_vX.X.X` as a base branch.

I don't think it's possible with our model to have a master as a base branch in a PR. Our intention is only to push changes to the master branch, not to make modifications directly on it.

ivadym: > That being said I don't think they should be modifying containers but as a sanity check?

We need to be able to track the old containers and label them with the release versions, that's the idea behind it.

.github/workflows/docker_build_publish.yml
ivadym: I would say yes, but what we include here appear later in the PR checks. A long name makes it difficult to access/read. But I agree with you, I can add some docstrings, that's a good idea 👍

.github/workflows/docker_build_publish_release.yml
ivadym: Yes good catch!

.github/workflows/docker_build_publish_release.yml
ivadym: This issue and PR was the reason: https://github.com/Clinical-Genomics/BALSAMIC/pull/1071

ivadym: I've also checked the original issue. It's still open and they are still recommending this workaround 🤷

docs/FAQs.rst
ivadym: Yes. But with this change at least we'll use in production the containers that we are testing during validation.

 I'm currently in the process of working through this, but my current conclusion is that in order to avoid the cache rebuilding, we must separate the workflow from the balsamic init process.

ivadym: Probably we'll have to go through one last refinement session with this 🤷

.github/workflows/docker_build_publish.yml
fevac: > But if we have hotfixes, the PR will have `release_vX.X.X` as a base branch.
> 
> I don't think it's possible with our model to have a master as a base branch in a PR. Our intention is only to push changes to the master branch, not to make modifications directly on it.

True. I guess this scenario would only be possible in the case that human errors happen and we do not name the branch release_XXX?

ivadym: But that's not the scope of this Github Action. Our goal here is to automatically push the container to docker hub from pull requests that involve changes to the containers folder, assigning a `develop` tag. We'll exclude the `release_X.X.X` branches, as they are handled by separate GitHub Actions.

The idea is to eliminate the need for manual builds during testing and we'll be able to finally use the develop containers for something.

Regarding the master branch. The only way that we'll have a PR with master as a base branch is probably if we want to open a PR to merge changes from master to develop, which is very unlikely.

Sorry I was not clear enough in the PR description, and this branch ignores/includes are quite confusing 🥵

ivadym: Ok, I understood what you mean. My implementation is not correct because of some incorrect assumptions I made regarding `on pull request`, nice that you pointed it out 🙇 🙇 🙇

ivadym: OK, I've fixed it. Very good call, thanks again!

khurrammaqbool (commented):
👍  
I think since it is a major structural change, this current container structure and the updated structure need to be described and compared along with the motivation to change in terms of the consequences of those changes. 
I have also some comments on general improvement of the documentation to reflect the changes and since we are addressing these changes, it will be nice to add the missing documentation for the whole process as well.

docs/FAQs.rst
```diff
@@ -66,33 +66,27 @@ Our release model looks like in the figure below:
...
+Release from `develop`:
+
+#. Switch to the master branch `git checkout master`.
+#. Pull the latest changes `git pull`.
+#. Switch to the `develop` branch: `git checkout develop`.
+#. Merge the latest changes from master into develop: `git merge master`.
+#. Resolve any conflicts, commit the changes, and push them to the develop branch.
+#. Create a new release branch: `git checkout -B release_vX.X.X` (adhere strictly to the naming convention for compatibility with container publishing GitHub Actions).
+#. Update the version number `X.X.X` in the `CHANGELOG.rst`.
+#. Open a pull request for the `release` branch, and after obtaining approval, merge it into `master`.
```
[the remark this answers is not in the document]
ivadym: I believe this two points are covered in Atlas: https://atlas.scilifelab.se/infrastructure/BALSAMIC/balsamic_gitmodel/

.github/workflows/docker_build_publish_release.yml
```diff
@@ -1,18 +1,15 @@
...
   main:
-    name: Docker image build test
+    name: Build and publish
+    if:  startsWith(github.head_ref, 'release_v')
     runs-on: ubuntu-20.04
     strategy:
-      fail-fast: false
+      fail-fast: true
       matrix:
         container-name: [align_qc, annotate, coverage_qc, varcall_cnvkit, varcall_py27, varcall_py3, balsamic, delly, vcf2cytosure, cnvpytor, somalier, ascatNgs]
```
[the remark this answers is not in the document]
ivadym: Let's address it also in a documentation PR as you suggested 📝

docs/FAQs.rst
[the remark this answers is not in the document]
ivadym: OK, sounds good you convinced me 👍

PR #1234, opened 2023-08-21
feat: add htslib container

khurrammaqbool:
### This PR:

Added: htslib container with samtools, bcftools, tabix and bgzip.

### Review and tests: 
- [x] Tests pass
- [ ] Code review

### DIFF
--- a/BALSAMIC/containers/htslib/Dockerfile
+++ b/BALSAMIC/containers/htslib/Dockerfile
@@ -11,15 +11,15 @@ LABEL about.documentation="http://www.htslib.org/doc/#manual-pages"
 LABEL about.license="MIT License (MIT)"
 
 RUN apt-get update && \
-     apt-get -y upgrade && \
-     apt-get -y  install --no-install-recommends \
-     samtools=1.13-4 bcftools=1.13-1 tabix=1.13+ds-2build1 && \
-     apt-get clean && \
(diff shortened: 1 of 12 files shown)


BALSAMIC/containers/htslib/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+LABEL software.version="1.13"
+LABEL about.summary="A unified C library for accessing common bioinformatics file formats"
+LABEL about.home="http://www.htslib.org/"
+LABEL about.documentation="http://www.htslib.org/doc/#manual-pages"
+LABEL about.license="MIT License (MIT)"
+
+RUN apt-get update && \
+     apt-get -y upgrade && \
+     apt-get -y  install --no-install-recommends \
+     samtools=1.13-4 bcftools=1.13-1 tabix=1.13+ds-2build1 && \
```
ivadym, line 16: Since we're saving these tools in a yaml file, couldn't we read it here to actually install them?

It would be easier to keep both of the files up to date

ivadym (approved):
Looks good! 👍 

Left you some minor comments 😊

BALSAMIC/containers/htslib/Dockerfile
khurrammaqbool: Let us take it up after all the containers are refactored.

PR #1242, opened 2023-09-06
refactor: reuse Balsamic CLI options

ivadym:
### Changed:
- Reuse common Balsamic CLI options

### Review and tests: 
- [x] Tests pass
  - [x] `balsamic config case` Screenshot 2023-09-08 at 16 37 25
  
  - [x] `balsamic config pon` Screenshot 2023-09-08 at 16 45 22
  
  - [x] `balsamic run analysis` Screenshot 2023-09-08 at 16 38 54

  - [x] `balsamic init` Screenshot 2023-09-08 at 16 47 43

- [ ] Code review


BALSAMIC/commands/options.py
```diff
@@ -1,93 +1,307 @@
...
+    type=click.Path(exists=True, resolve_path=True),
+    required=False,
+    help="Panel of normal reference (.cnn) for CNVkit",
+)
+
+OPTION_PON_VERSION = click.option(
+    "-v",
+    "--version",
+    default="v1",
+    type=str,
```
mathiasbio, line 286: ? 🤪 
```suggestion
    type=click.STRING,
```

BALSAMIC/commands/options.py
```diff
@@ -98,11 +312,94 @@
...
+    type=click.Path(exists=True, resolve_path=True),
+    required=False,
+    help="VCF path of Swegen SNV frequency database (WGS analysis workflow)",
+)
+
+OPTION_SWEGEN_SV = click.option(
+    "--swegen-sv",
+    type=click.Path(exists=True, resolve_path=True),
+    required=False,
+    help="VCF path of Swegen SV frequency database (WGS analysis workflow)",
```
mathiasbio, line 381: I think Swegen is used in TGA as well 🤔 

```suggestion
    help="VCF path of Swegen SV frequency database",
```

BALSAMIC/commands/options.py
```diff
@@ -98,11 +312,94 @@
...
+    "--snakemake-opt",
+    multiple=True,
+    help="Options to be passed to Snakemake",
+)
+
+OPTION_SWEGEN_SNV = click.option(
+    "--swegen-snv",
+    type=click.Path(exists=True, resolve_path=True),
+    required=False,
+    help="VCF path of Swegen SNV frequency database (WGS analysis workflow)",
```
mathiasbio, line 374: I think Swegen is used in TGA as well 🤔  🤔 
```suggestion
    help="VCF path of Swegen SNV frequency database",
```

BALSAMIC/commands/report/deliver.py
```diff
@@ -22,50 +29,21 @@
...
-    show_default=True,
-    help="a: append rules-to-deliver to current delivery options. "
-    "r: reset current rules to delivery to only the ones specified",
-)
-@click.option(
-    "--disable-variant-caller",
-    help=f"Run workflow with selected variant caller(s) disable. Use comma to remove multiple variant callers. Valid "
-    f"values are: {list(VCF_DICT.keys())}",
-)
+@click.command("deliver", short_help="Creates a JSON report file with output files")
```
mathiasbio, line 32: Is this a typo or has it changed from a YAML to a JSON? 👀 

```suggestion
@click.command("deliver", short_help="Creates a YAML report file with output files")
```

mathiasbio (approved):
I don't have many comments in the code beyond those few I already made : ) I'm trying to think of how this will affect developing in balsamic. 

It might make it slightly, slightly more complicated if you want to understand how a cli argument is defined, you just have to go the extra step and look in options. But it's so minor and I think we can get used to it.

On the positive side we have the benefit of not needing to update more command-line interfaces than one when we need to make changes. I also like that it makes it clearer that what we are expecting for the different commands, like PON and case, is the same argument, for instance for case and fastq-dir. 

Yes...I like it! 🌟

BALSAMIC/snakemake_rules/align/sentieon_alignment.rule
```diff
@@ -101,7 +101,7 @@ rule sentieon_realign:
         Path(benchmark_dir, "sentieon_realign_{sample_type}.{sample}.tsv").as_posix()
     params:
         tmpdir = tempfile.mkdtemp(prefix=tmp_dir),
-        sentieon_exec = config["SENTIEON_EXEC"],
+         sentieon_exec = config["SENTIEON_EXEC"],
```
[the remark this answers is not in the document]
mathiasbio: Best code-review comment so far I think.

BALSAMIC/commands/options.py
ivadym: 💘

BALSAMIC/commands/options.py
ivadym: True! I think we forgot to remove it when we also added it to TGA. Thanks!

BALSAMIC/commands/report/deliver.py
ivadym: The .hk files are actually JSONs. But it's true that we also generate a YAML file. I'll make it more general 👍

BALSAMIC/snakemake_rules/align/sentieon_alignment.rule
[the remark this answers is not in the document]
ivadym: I'm proud 🧠

ivadym:
> I don't have many comments in the code beyond those few I already made : ) I'm trying to think of how this will affect developing in balsamic.
> 
> It might make it slightly, slightly more complicated if you want to understand how a cli argument is defined, you just have to go the extra step and look in options. But it's so minor and I think we can get used to it.
> 
> On the positive side we have the benefit of not needing to update more command-line interfaces than one when we need to make changes. I also like that it makes it clearer that what we are expecting for the different commands, like PON and case, is the same argument, for instance for case and fastq-dir.
> 
> Yes...I like it! 🌟

Nice! I think keeping it at one place overall will make it more maintainable and easy to change and understand. Thanks for the review ⭐

PR #1244, opened 2023-09-07
feat: Allow custom cache version option for analysis

ivadym:
### Changed:
- Add cache version option to config case

### Review and tests: 
- [x] Tests pass
  - [x] `balsamic config case --cache-version` Screenshot 2023-09-08 at 16 28 28
- [ ] Code review


mathiasbio (approved):
So for most use-cases this doesn't change anything as the cache-version is by default the current balsamic version. But this allows us, in combination with the changes to the init to build a cache with the release-version we intend to use, and then despite not having bumped the version we can choose the later version for the validation. 🤔 sounds really good! I also found no issues with the code except one very very serious one...(just kidding)
Super nice job! 🌟 🧙 🌟

PR #1251, opened 2023-09-13
refactor: Update reference.json to use relative paths

ivadym:
### This PR:

In order to be able to copy the balsamic cache from stage to production, we need to start saving the cache reference file links as relative paths (`reference.json`).

To maintain simplicity, the `config.json` reference file will continue to point to stage files, as the cache will be built primarily in the stage from this point onward.

### Changed:
- Update `reference.json` file to use relative paths

### Review and tests: 
- [x] Tests pass
  - [x] `balsamic init`. Saves relative paths.
  Screenshot 2023-09-13 at 10 48 47

  - [x] `balsamic config case`. Reads relative paths, but saves them as absolute.
  Screenshot 2023-09-13 at 10 53 02

- [ ] Code review


BALSAMIC/commands/config/pon.py
```diff
@@ -64,16 +67,14 @@ def pon_config(
     umi_trim_length: bool,
     version: str,
 ):
-    reference_config = os.path.join(
-        balsamic_cache, cache_version, genome_version, "reference.json"
+    references_path: Path = Path(balsamic_cache, cache_version, genome_version)
+    references: Dict[str, Path] = get_absolute_paths_dict(
```
mathiasbio, line 71: Nice solution! : )

tests/conftest.py
```diff
@@ -484,24 +485,27 @@ def no_write_perm_path(tmp_path_factory) -> str:
...
-    cache_reference_json.touch()
-    with open(cache_reference_json, "w") as fp:
-        json.dump(reference, fp)
-
-    return cache_dir.as_posix()
+@pytest.fixture(scope="session")
+def balsamic_cache(
+    tmp_path_factory: TempPathFactory, reference: Dict[str, Path], references_dir: Path
+) -> str:
+    """Create and return the path for balsamic-cache"""
```
mathiasbio, line 497: Geez, so sloppy... 🙄 
```suggestion
    """Create and return the path for balsamic-cache."""
```

tests/conftest.py
```diff
@@ -484,24 +485,27 @@ def no_write_perm_path(tmp_path_factory) -> str:
...
+    balsamic_cache: Path = tmp_path_factory.mktemp("balsmic_cache")
```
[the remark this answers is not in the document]
ivadym: I forgot to change it at the end 😆 , thanks!

PR #1252, opened 2023-09-13
feat: add cnvkit container

khurrammaqbool:
### This PR:

Added: cnvkit container.

### Review and tests: 
- [x] Tests pass
- [x] Code review


ivadym (approved):
Nice 🏆 

Given that this is an update of the `varcall_cnvkit` container, should we consider eliminating it, or is this intended for a future PR?

By the way, let's make sure to include all these new containers in the DockerContainers constant 🆙

khurrammaqbool:
Thanks for the review! 

> Given that this is an update of the `varcall_cnvkit` container, should we consider eliminating it, or is this intended for a future PR?

It is part of `varcall_cnvkit`
 
> By the way, let's make sure to include all these new containers in the `DockerContainers` constant 🆙

I am aware of it and did not add yet because as I mentioned above it is only part of `varcall_cnvkit`.

Issue #1271, opened 2023-10-02
VarDict bug after realignment


mathiasbio:
**Describe the bug**

This error has been thrown quite randomly by VarDict and causes a crash. After restarting it succeeds. So it seems quite random.
I saw it previously in this PR https://github.com/Clinical-Genomics/BALSAMIC/pull/1176 but misinterpreted the error. Looking at it today it seems to be related to the realignment which was introduced for TGA in this PR, and which has been mentioned in this issue on VarDict as being related to realignment: https://github.com/AstraZeneca-NGS/VarDictJava/issues/341

```
There was Exception while processing variant on 70121684 on region 17:70119553-70122691. The processing will be continued from the next variant.
java.lang.IllegalArgumentException
	at java.nio.Buffer.position(Buffer.java:244)
	at htsjdk.samtools.MemoryMappedFileBuffer.skipBytes(MemoryMappedFileBuffer.java:49)
```

I think we short revert the TGA workflow to not do realignment as this is what we were doing before, and since we have already discussed removing realignment for all workflows, this seems like the safest and easiest choice. We can keep it for WGS for now.

**To Reproduce**

Seems quite random, and is therefore difficult to reproduce reliably.

**Expected behavior**

**If workflow, which rules**
If possible, and using the Snakemake workflows, the name of the affected rules and workflows.

**Screenshots**

**Version (please complete the following information):**
`balsamic --version` development branch (2023-10-02)

**Additional context**

mathiasbio:
Solved with: https://github.com/Clinical-Genomics/BALSAMIC/pull/1272

PR #1272, opened 2023-10-02
fix: Remove local realignment for TGA

mathiasbio:
### This PR:

This PR attempts to solve this bug: https://github.com/Clinical-Genomics/BALSAMIC/issues/1271 by removing realignment step for TGA. 

Removed: Realignment for TGA workflows.

unitedbeagle DAG:
image

civilsole DAG:
image

Running tests to verify integrity of workflows. Check boxes when workflow finishes without errors:
- [x] TGA T/N case unitedbeagle
- [x] WGS T case civilsole

### Review and tests: 
- [x] Tests pass
- [x] Code review


BALSAMIC/snakemake_rules/align/bam_compress.rule
```diff
@@ -46,3 +46,49 @@ rule bam_compress_normal:
...
+        housekeeper_id= {"id": "{sample}", "tags": "normal"}
+    threads:
+        get_threads(cluster_config, "bam_compress")
+    message:
+        "Compressing bam to cram for {params.sample_id}"
+    shell:
+        """
+samtools view -h -T {input.fasta} --threads {threads} -C -o {output.cram} {input.bam}; 
+samtools index {output.cram}; 
+        """
```
ivadym, line 94: EOF

BALSAMIC/snakemake_rules/align/bam_compress.rule
```diff
@@ -46,3 +46,49 @@ rule bam_compress_normal:
 samtools view -h -T {input.fasta} --threads {threads} -C -o {output.cram} {input.bam}; 
 samtools index {output.cram}; 
         """
+
+rule bam_compress_tumor_wgs:
```
ivadym, line 50: I'm not sure which is the best strategy here: defining an input file variable depending on the analysis type (WGS vs. targeted) or having code duplication in these rules. I would say that an if statement is better, mainly for maintainability, but feel free to disagree.

ivadym (approved):
Looks good 👍 🥇 

Left you just a minor comment 🧂

BALSAMIC/snakemake_rules/align/bam_compress.rule
mathiasbio: No it's a good idea! I didn't think for very long before making this change. I think I have a better and pretty obvious solution now :) Just waiting for the test-runs to finish!

ivadym: Nice ⭐

PR #1284, opened 2023-10-16
updated python requirement in install.rst

JiayangZhou:
### This PR:
Document update

### Reason for this PR:

- according to https://github.com/Clinical-Genomics/BALSAMIC/blob/master/docs/install.rst#step-1-installing-balsamic the python version recommended with conda is 3.7 but in the software requirement - session it says 3.6, this is the first place there is contradiction.
- most importantly, the evidence for 3.6 is not supported at all, is in setup.py https://github.com/Clinical-Genomics/BALSAMIC/blob/e85f45b84e20564c798ee5d8f58deb8086ee5c4f/setup.py#L13 the click version required is 8.1.3, but python3.6 only support click up to  8.0.4, if you install python3.6 and will not be able to upgrade click higher than 8.0.4 so this part of document is misleading 

### Review and tests: 
- [ ] Tests pass
- [ ] Code review

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -11,7 +11,7 @@ Software Requirements
 
 - Conda >=version 4.5.0: For detailed software and python requirements please see ``setup.py`` and ``BALSAMIC/conda/balsamic.yaml``
 - Singularity >=version 3.0.0: BALSAMIC uses singularity to run vairous parts of the workflow. 
-- Python 3.6
+- Python 3.7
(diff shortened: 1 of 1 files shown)


ivadym:
Thank you for your interest in Balsamic!

I would like to direct you to our [documentation](https://balsamic.readthedocs.io/en/latest/FAQs.html#git-related-questions) regarding git contributions, as we follow a [gitflow model](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow). If you'd like to make contributions, please use the `develop` branch as your base branch.

Regarding this PR, we've recently migrated to Python 3.11, and the documentation has been updated accordingly in our `develop` branch. We're also planning to make a new release (v13) in the upcoming weeks. But thanks for pointing it out in any case!

PR #1287, opened 2023-10-17
Added ability to reuse cache dir

JiayangZhou:
### This PR:

Added: for new features.

To Reproduce Error. If you chose a balsamic cache directory that was previously used (the reason you reuse a cache directory could be the old container pull was "interrupted" for some reason and you want to continue from there without starting over). With the current code, It will error out because snakemake is putting a lock to that directory. And the error is suggesting you can use --unlock but this flag is not accepted by balsamic. Thus adding it to be able to utilize --unlockflag from snakemake

### Review and tests: 
- [ ] Tests pass
- [ ] Code review

### DIFF
--- a/BALSAMIC/commands/init/base.py
+++ b/BALSAMIC/commands/init/base.py
@@ -170,6 +170,13 @@ LOG = logging.getLogger(__name__)
     is_flag=True,
     help="Instruct snakemake to be quiet! No output will be printed",
 )
+@click.option(
+    "-u",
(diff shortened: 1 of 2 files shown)


ivadym:
For these scenarios, we typically use the `snakemake_opt` flag in Balsamic. You can specify multiple options there, and they will be appended to the Snakemake command:

```
--snakemake-opt "--unlock --cores 24"
```

Issue #1240, opened 2023-08-31
Calculate GC/AT dropout for WGS cases


pbiology:
## Need
It would be good to start tracking GC and AT dropout for all WGS cases. Right now we are just doing it for panels.

## Suggested approach
Run picard hs metrics also on WGS cases. Preferably with the same exome BED file as used in rare disease. Could be worthwhile asking the RD team if we should update which file is used for the exome.

## Considered alternatives
Could we otherwise somehow get the whole genome, and not only exome?

## Can be closed when
- [ ] BALSAMIC captures GC and At dropout in the multiqc.json file

pbiology:
Let's implement as in MIP for release 13, and then for next release we should re-evaluate exactly which regions should be underlying the GC/AT dropout calculations

mathiasbio:
RD seems to be using an old bedfile from the twist exome prep: "twistexomerefseq_9.1_hg19_design.bed.pad100.interval_list" I wonder if it isn't simply better for us that we use the RefGene bedfile that we're already using in balsamic as it seems to be a more general, untampered file.

It would also be easier to implement this if we were using that file as implementing the same as RD is using would require us to either:
- add another reference file to download in our init, and then do some modification of it to get the "pad100" (as I don't think this file is available online)
- alternatively to save this exome-bed in a reference folder and add it as an argument to balsamic config case

To start I will run with the exome-fil we have, AND the RD one, and compare the stats we get. If they are similar, which I think they will be as the regions included in these files are so large it should converge on the same value, then I think we can just run with the RefGene one. I'll write the results here.

pbiology:
Sounds like a good way forward. We should not skip on aligning with the RD group before making any decisions. Having the values comparable seems quite valuable

mathiasbio:
[...]

PR #1288, opened 2023-10-17
feat: add extra qc metrics

mathiasbio:
### This PR:

The aim of this PR is primarily to adds GC dropout to WGS by adding picard hsmetrics to the WGS workflow, using the refgene bedfile. 

See the issue here where a discussion was had regarding which bed file to use to calculate GC_dropout in WGS (to align with the RD pipeline, or to use the RefGene bedfile already in use in balsamic) (https://github.com/Clinical-Genomics/BALSAMIC/issues/1240).

It also:
- adds picard_CollectGcBiasMetrics from Picard (since this is a GC bias metric that is based on looking at the entire genome, it could be an interesting metric to track specifically for WGS)
- corrects some QC output names from TGA which still retained the realign part of their output names after the PR that removed realignment for TGA. (https://github.com/Clinical-Genomics/BALSAMIC/pull/1272)

Added: 
- picard_HsMetrics for WGS (using refgene as a target bedfile)
- picard_CollectGcBiasMetrics for WGS

Fixed: 
- corrected names of QC metrics output-files for TGA to reflect the bamfile used there (not realigned, only deduped). 

### PR specific checks

Integrity:
- [x] TGA T/N workflow is still intact after renaming
- [x] TGA T/N multiqc_data.json contains the same info as in develop
- [x] WGS T-only workflow finishes without errors

New QC-metrics:
- [x] GC_dropout metrics is added to WGS multiqc_data.json
- [x] additional GC_bias metrics is added to WGS multiqc_data.json

### Review and tests: 
- [x] Tests pass
- [ ] Code review


BALSAMIC/snakemake_rules/quality_control/picard_wgs.rule
```diff
@@ -7,6 +7,84 @@ picard_metrics_wildcard = ["alignment_summary_metrics", "base_distribution_by_cy
...
+rule picard_CollectHsMetrics_WGS:
+    input:
+        fadict = (config_model.reference["reference_genome"]).replace(".fasta",".dict"),
+        bed = config_model.reference["refgene_bed"],
+        bam = lambda wildcards: config_model.get_final_bam_name(bam_dir = bam_dir, sample_name = wildcards.sample),
+        fa = config_model.reference["reference_genome"],
+    output:
+        hs_metrics = qc_dir + "{sample}.dedup.realign.hsmetric"
+    benchmark:
+        Path(benchmark_dir + "picard_CollectHsMetrics_" + "{sample}.tsv").as_posix()
```
ivadym, line 19:
```suggestion
        Path(benchmark_dir, "picard_CollectHsMetrics_{sample}.tsv").as_posix()
```

BALSAMIC/snakemake_rules/quality_control/picard_wgs.rule
```diff
@@ -7,6 +7,84 @@ picard_metrics_wildcard = ["alignment_summary_metrics", "base_distribution_by_cy
...
+        tmpdir = tempfile.mkdtemp(prefix=tmp_dir),
+        bed_name = os.path.basename(config_model.reference["refgene_bed"]),
+        sample = '{sample}'
+    threads:
+        get_threads(cluster_config, 'picard_CollectHsMetrics')
+    message:
+        "Calculating picard HsMetrics for sample '{params.sample}'"
+    shell:
+        """
+mkdir -p {params.tmpdir};
```
ivadym, line 33:
```suggestion
```

BALSAMIC/snakemake_rules/quality_control/picard_wgs.rule
```diff
@@ -7,6 +7,84 @@ picard_metrics_wildcard = ["alignment_summary_metrics", "base_distribution_by_cy
...
+    output:
+        hs_metrics = qc_dir + "{sample}.dedup.realign.hsmetric"
+    benchmark:
+        Path(benchmark_dir + "picard_CollectHsMetrics_" + "{sample}.tsv").as_posix()
+    singularity:
+        Path(singularity_image, config["bioinfo_tools"].get("picard") + ".sif").as_posix()
+    params:
+        mem = "16g",
+        tmpdir = tempfile.mkdtemp(prefix=tmp_dir),
+        bed_name = os.path.basename(config_model.reference["refgene_bed"]),
```
ivadym, line 25: lets use pathlib 😄

BALSAMIC/snakemake_rules/quality_control/picard_wgs.rule
```diff
@@ -7,6 +7,84 @@ picard_metrics_wildcard = ["alignment_summary_metrics", "base_distribution_by_cy
...
+        sample = '{sample}'
```
ivadym, line 26: And also double quotes when possible, please review this file 💌 

```suggestion
        sample = "{sample}"
```

BALSAMIC/snakemake_rules/quality_control/picard_wgs.rule
```diff
@@ -7,6 +7,84 @@ picard_metrics_wildcard = ["alignment_summary_metrics", "base_distribution_by_cy
...
+
+picard -Djava.io.tmpdir={params.tmpdir} -Xmx{params.mem} \
+CollectHsMetrics \
+BI={input.bam}.picard.bedintervals \
+TI={input.bam}.picard.bedintervals \
+I={input.bam} \
+O={output.hs_metrics} \
+R={input.fa} \
+BAIT_SET_NAME={params.bed_name} \
+COVERAGE_CAP=50000 \
```
ivadym, line 50: We should start considering defining these constants  🤔 Maybe out of scope here since we have to address it everywhere

BALSAMIC/snakemake_rules/quality_control/picard_wgs.rule
```diff
@@ -7,6 +7,84 @@ picard_metrics_wildcard = ["alignment_summary_metrics", "base_distribution_by_cy
...
+        tmpdir = tempfile.mkdtemp(prefix=tmp_dir),
+        bed_name = os.path.basename(config_model.reference["refgene_bed"]),
+        sample = '{sample}'
+    threads:
+        get_threads(cluster_config, 'picard_CollectHsMetrics')
+    message:
+        "Calculating picard HsMetrics for sample '{params.sample}'"
+    shell:
+        """
+mkdir -p {params.tmpdir};
```
ivadym, line 77:
```suggestion
```

BALSAMIC/snakemake_rules/quality_control/picard_wgs.rule
```diff
@@ -7,6 +7,84 @@ picard_metrics_wildcard = ["alignment_summary_metrics", "base_distribution_by_cy
...
+rule picard_CollectGcBiasMetrics:
+    input:
+        bam = lambda wildcards: config_model.get_final_bam_name(bam_dir = bam_dir, sample_name = wildcards.sample),
+        fa = config_model.reference["reference_genome"],
+    output:
+        gc_bias_metrics = qc_dir + "{sample}.dedup.realign.gc_bias_metrics",
+        gc_bias_sum_metrics = qc_dir + "{sample}.dedup.realign.gc_bias_metrics.sum_metrics",
+        gc_bias_sum_chart = qc_dir + "{sample}.dedup.realign.gc_bias_metrics.chart"
+    benchmark:
+        Path(benchmark_dir + "picard_CollectHsMetrics_" + "{sample}.tsv").as_posix()
```
ivadym, line 63:
```suggestion
        Path(benchmark_dir, "picard_CollectHsMetrics_{sample}.tsv").as_posix()
```

BALSAMIC/snakemake_rules/quality_control/picard_wgs.rule
```diff
@@ -7,6 +7,84 @@ picard_metrics_wildcard = ["alignment_summary_metrics", "base_distribution_by_cy
...
+    singularity:
+        Path(singularity_image, config["bioinfo_tools"].get("picard") + ".sif").as_posix()
+    params:
+        mem = "16g",
+        tmpdir = tempfile.mkdtemp(prefix=tmp_dir),
+        bed_name = os.path.basename(config_model.reference["refgene_bed"]),
```
ivadym, line 69: pathlib

BALSAMIC/snakemake_rules/quality_control/picard_wgs.rule
```diff
@@ -7,6 +7,84 @@ picard_metrics_wildcard = ["alignment_summary_metrics", "base_distribution_by_cy
...
+mkdir -p {params.tmpdir};
+export TMPDIR={params.tmpdir};
+
+picard -Djava.io.tmpdir={params.tmpdir} -Xmx{params.mem} \
+CollectGcBiasMetrics \
+I={input.bam} \
+O={output.gc_bias_metrics} \
+R={input.fa} \
+S={output.gc_bias_sum_metrics} \
+CHART={output.gc_bias_sum_chart} ;
```
ivadym, line 86:
```suggestion
CHART={output.gc_bias_sum_chart};
```

BALSAMIC/snakemake_rules/quality_control/picard_wgs.rule
```diff
@@ -7,6 +7,84 @@ picard_metrics_wildcard = ["alignment_summary_metrics", "base_distribution_by_cy
...
+      """
+
+rule picard_CollectGcBiasMetrics:
+    input:
+        bam = lambda wildcards: config_model.get_final_bam_name(bam_dir = bam_dir, sample_name = wildcards.sample),
+        fa = config_model.reference["reference_genome"],
+    output:
+        gc_bias_metrics = qc_dir + "{sample}.dedup.realign.gc_bias_metrics",
+        gc_bias_sum_metrics = qc_dir + "{sample}.dedup.realign.gc_bias_metrics.sum_metrics",
+        gc_bias_sum_chart = qc_dir + "{sample}.dedup.realign.gc_bias_metrics.chart"
```
ivadym, line 61: Isn't the produced file a PDF?
```suggestion
        gc_bias_sum_chart = qc_dir + "{sample}.dedup.realign.gc_bias_metrics.chart.pdf"
```

BALSAMIC/snakemake_rules/quality_control/picard_wgs.rule
```diff
@@ -7,6 +7,84 @@ picard_metrics_wildcard = ["alignment_summary_metrics", "base_distribution_by_cy
...
+COVERAGE_CAP=50000 \
+METRIC_ACCUMULATION_LEVEL=ALL_READS;
+      """
+
+rule picard_CollectGcBiasMetrics:
+    input:
+        bam = lambda wildcards: config_model.get_final_bam_name(bam_dir = bam_dir, sample_name = wildcards.sample),
+        fa = config_model.reference["reference_genome"],
+    output:
+        gc_bias_metrics = qc_dir + "{sample}.dedup.realign.gc_bias_metrics",
```
ivadym, line 59:
```suggestion
        gc_bias_metrics = qc_dir + "{sample}.dedup.realign.gc_bias_metrics.txt",
```

BALSAMIC/snakemake_rules/quality_control/picard_wgs.rule
```diff
@@ -7,6 +7,84 @@ picard_metrics_wildcard = ["alignment_summary_metrics", "base_distribution_by_cy
...
+        gc_bias_sum_metrics = qc_dir + "{sample}.dedup.realign.gc_bias_metrics.sum_metrics",
```
ivadym, line 60:
```suggestion
        gc_bias_summary_metrics = qc_dir + "{sample}.dedup.realign.gc_bias_metrics.summary_metrics.txt",
```

BALSAMIC/snakemake_rules/quality_control/picard_wgs.rule
```diff
@@ -7,6 +7,84 @@ picard_metrics_wildcard = ["alignment_summary_metrics", "base_distribution_by_cy
...
 
 
+rule picard_CollectHsMetrics_WGS:
+    input:
+        fadict = (config_model.reference["reference_genome"]).replace(".fasta",".dict"),
+        bed = config_model.reference["refgene_bed"],
+        bam = lambda wildcards: config_model.get_final_bam_name(bam_dir = bam_dir, sample_name = wildcards.sample),
+        fa = config_model.reference["reference_genome"],
+    output:
+        hs_metrics = qc_dir + "{sample}.dedup.realign.hsmetric"
```
ivadym, line 17: I'm not sure why we've been using this naming strategy here, but maybe we should start giving these files a proper suffix?
```suggestion
        hs_metrics = qc_dir + "{sample}.dedup.realign.hsmetric.txt"
```

ivadym (commented):
It looks good 🥇 All the new picard files are fed to Multiqc and the rule arguments look fine 💯

BALSAMIC/snakemake_rules/quality_control/picard_wgs.rule
mathiasbio: Added a new issue to refactor these type of parameters into constants:
https://github.com/Clinical-Genomics/BALSAMIC/issues/1301

BALSAMIC/snakemake_rules/quality_control/picard_wgs.rule
mathiasbio: I added it everywhere I could find, except for the "picard_CollectMultipleMetrics" outputs since it seems there that it is Picard that decides the full final name and we have no control over the suffix : /

ivadym (approved):
Fantastic work!!! ⭐ 👨‍💻 ⭐ 

We should add the relevant metrics to the `metrics_deliverables.yaml` file. Probably a good place will be during the QC review: https://github.com/Clinical-Genomics/BALSAMIC/issues/1297

pbiology:
So having talked a bit to @ivadym, @fevac and @jemten, perhaps we should make an effort to keep the current QC calculations as similar as possible between the cancer and RD pipelines.  
I think there is consensus that using a static file, such as RefGene is the way forward, but in order to make things similar right now, we should perhaps start using the twistExome BED file used in MIP also in BALSAMIC, and then we have a quick look for which static BED file would be the best replacement, and we implement it in both pipelines at a future date.

mathiasbio:
@pbiology I can see the appeal of using the same file as it more or less unifies this metric for all WGS samples. I agree with the image that it would be nice to just look at all WGS samples, as data gathered from the same lab-method, and to have a unified may of getting these QC-stats. 

Practically implementing this change means adding another command-line argument to config a case, and the associated change in CG, which will be quite minor, and some new logic in balsamic to maybe allow it to be optional (to not force researchers to create these interval_list files). Changes should be fairly quick however. 

Part of me hurts a little needing to implement logic in balsamic to make 1 metric unified with another pipeline 😅 probably there's a quite a few more values that would be different for the same sample if we ran them in MIP or Balsamic. And I guess for the future we should continue in this spirit of having identical QC workflows for RD and cancer, or maybe make a separate workflow just for QC. 

For the purposes of trending I don't think it matters that we have slightly differently skewed GC_dropout for the different pipelines since they should both reflect some underlying truth of the lab-prep-quality regardless. Maybe it could even be good to get some different perspectives on the same stat 🤷 (as long as there's no obvious optimal way of calculating them)

But if we are using the same metric I can see how we can sort of steal some knowledge from the MIP metrics if we want to implement some QC-threshold for WGS in balsamic, if we use RefSeq we may need to build up this knowledge with time, but maybe if we use the same as MIP we can just learn from their stats. Maybe that's the most important point?

pbiology:
> For the purposes of trending I don't think it matters that we have slightly differently skewed GC\_dropout for the different pipelines since they should both reflect some underlying truth of the lab-prep-quality regardless. Maybe it could even be good to get some different perspectives on the same stat 🤷 (as long as there's no obvious optimal way of calculating them)

I think this is the biggest for me. Looking at the data you presented in #1240 , not only was the GC/AT\_Drop different, but also various coverage values were quite different depending on the BED file used. It is quite appealing to say that if we look at WGS data, and we use Picards caluclate hs mtrics in both, we should be able to fully compare values comping from two different pipelines.

mathiasbio:
But if what we're caring about is trending these QC-metrics from the same method, to be able to monitor the quality of the prep, I think we're already getting most of that from the MIP WGS-cases which have recorded this metric for a long time.

The extra stats from Balsamic just seems like more data-points to me, and at that point I don't see that it matters so much if these data-points are skewed in any particular way based on checking different regions, and if we're already planning to change to a more general bed-file like RefSeq maybe it's better that we've already starting trending on this bedfile in parallel? 

What we will say if we collect the data from balsamic and MIP are things like "GC_DROPOUT in balsamic is consistently [5% higher/lower] than in MIP due to using different bedfiles, but the trend is the same..." I don't know if it's worth it to add new logic for this

mathiasbio:
I don't think I would have time to make the change to twist before validation on Monday. Especially since it also requires making it work for hg38 which is not an issue with the RefGene bed, and that there are still some tasks left over in CG to solve before release. If everyone is ok with it, I'm going to merge this tomorrow morning 🔥 🧙 🔥

Issue #1285, opened 2023-10-17
Develop Branch: Unable to Store Cases


ivadym:
**Describe the bug**

It's not possible to store any cases in Housekeeper. This issue affects all workflows.

**To Reproduce**
- `/home/proj/stage/bin/miniconda3/envs/S_hermes/bin/hermes convert deliverables ...`
- `cg workflow <workflow> store <case>`

**Expected behavior**
- Cases should be successfully stored in Housekeeper.

**If workflow, which rules**
- `balsamic-qc`
> balsamic hasta.scilifelab.se 2023-10-17 10:33:14 snakemake.logging pid:226536 [ERROR] AttributeError in file /home/proj/stage/bin/miniconda3/envs/S_balsamic/lib/python3.11/site-packages/BALSAMIC/workflows/QC.smk, line 136:
> 'list' object has no attribute 'keys'

- `balsamic` (WGS, TGA) and `balsamic-umi` 

> 2023-10-17 10:37:37 hasta.scilifelab.se cg_hermes.deliverables[233561] INFO Parsing deliverables for balsamic
> 2023-10-17 10:37:37 hasta.scilifelab.se cg_hermes.deliverables[233561] WARNING Mandatory file (tumor-cram, cram) is missing
> 2023-10-17 10:37:37 hasta.scilifelab.se cg_hermes.deliverables[233561] WARNING Mandatory file (tumor-cram-index, cram) is missing
> 2023-10-17 10:37:37 hasta.scilifelab.se cg_hermes.cli.convert[233561] ERROR Deliverables is missing mandatory files

**Proposed solution**
- [x] `balsamic-qc`: adapt QC.smk to the new sample structure (list, not dict) (https://github.com/Clinical-Genomics/BALSAMIC/pull/1295)
- [x] `balsamic`: include cram files in the HK deliverable file (https://github.com/Clinical-Genomics/BALSAMIC/pull/1307)

**Version (please complete the following information):**
`balsamic --version` `develop`

ivadym:
Addressed!

PR #1295, opened 2023-10-24
fix: Store QC workflow

ivadym:
### Fixed:
- Store for QC workflow

### Review and tests: 
- [x] Tests pass ✅ 
  - [x] `cg workflow balsamic-qc store <case-id>`
- [ ] Code review


BALSAMIC/workflows/PON.smk
```diff
@@ -84,8 +84,6 @@ if not Path(config["SENTIEON_EXEC"]).exists():
 
 panel_name = os.path.split(target_bed)[1].replace('.bed','')
 
-coverage_references = expand(cnv_dir + "{sample}.{cov}coverage.cnn", sample=config["samples"], cov=['target','antitarget'])
-baited_beds = expand(cnv_dir + "{cov}.bed", cov=['target','antitarget'])
```
mathiasbio, line 88: Ah nice catch, we don't need these as they are inputs to the final rule in the CNVkit pon creation workflow!

mathiasbio (approved):
Looks good! 🌴 🧙 🔥  Is there anywhere that I can read more about why this fix was required? If there is maybe you can link it here? ❤️

ivadym:
> Looks good! 🌴 🧙 🔥 Is there anywhere that I can read more about why this fix was required? If there is maybe you can link it here? ❤️

This PR is linked to it's respective issue 😜

https://github.com/Clinical-Genomics/BALSAMIC/issues/1285

mathiasbio:
😬 
...
It was a test, you have passed... 🧙

PR #1296, opened 2023-10-24
refactor: Remove archived/outdated workflows and scripts

ivadym:
### This PR:

Having dead code in the project that keeps popping out when I do global searches is annoying. I'm removing it.

### Removed:
* Archived/outdated workflows and scripts

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


mathiasbio (approved):
Those scripts look like they could be useful, or have been useful at some point. But we can dig them out of the grave in the future (💀 🧙 💚 ) if we realise we need to make coverage-plots from sambamba, or plot mutational signatures, or want to...I don't know exactly what VariantReport was used for...filtering variants and calculating TMB? 🔥

PR #1303, opened 2023-10-27
refactor: Use StrEnum from python enum 

ivadym:
### This PR:

After upgrading to Python 3.11, we can now use `StrEnum` directly from the `enum` module. There's no longer a need to access` .value`, as we can work with these constants directly.

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


mathiasbio (approved):
Nice 🧙 🔥  I can find nothing wrong with this! I stamp this! ✉️

PR #1308, opened 2023-11-01
chore: Update snakemake version

ivadym:
### This PR:

Updating Snakemake version to the latest before release.

### Review and tests: 
- [x] Tests pass
- [ ] Code review


setup.py
```diff
@@ -97,7 +97,7 @@
     "six==1.16.0",
     "smart-open==6.3.0",
     "smmap==5.0.0",
-    "snakemake==7.32.0",
+    "snakemake==7.32.4",
```
mathiasbio, line 100: Do you think this will impact the key_error issue? :)

mathiasbio (approved):
Nice! ⭐ ❤️ ⭐

setup.py
ivadym: I don't think so, they fixed some things related the key error, use rules, ... but I don't think it will affect us. It's always nice to keep it up to date though :)

PR #1323, opened 2023-11-08
fix: CADD annotations container path & (some) code smells

ivadym:
### This PR:
String has been assigned instead of a Path for CADD annotations. There are also some code smells that we overlooked. 

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


mathiasbio (approved):
Amazing, thank you <3 you should sell this sweet smelling code as a perfume 🌷

PR #1326, opened 2023-11-13
feat: Build and push release containers only when release PR is open or containers modified

ivadym:
### This PR:

We should build and push release containers only when the release candidate PR is opened or when container fixes are pushed to it. With this approach, we wouldn't need to rebuild the cache with every fix to the release branch.

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


ivadym:
Looking it more closely, it would requiere manual triggering to make it work. If it becomes a problem could be brought for refinement.

PR #1329, opened 2023-11-15
fix: Vardict Java exceptions

ivadym:
### Review and tests: 
- [ ] Tests pass
- [ ] Code review


ivadym:
Doesn't seem to solve the problem, will keep looking into it

PR #1332, opened 2023-11-27
Add memory to vardict

mathiasbio:
### This PR:

(If this is a release PR, no need to add following. Leave this part empty)
(Use the following lines to create a PR text body. Make sure to remove all non-relevant one after you're done)
(Repeat each field as many times as necessary)

Added: for new features.
Changed: for changes in existing functionality.
Deprecated: for soon-to-be removed features.
Fixed: for any bug fixes.
Security: in case of vulnerabilities.

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


BALSAMIC/snakemake_rules/align/postprocess_bam.rule
```diff
@@ -0,0 +1,47 @@
...
+
+samtools view --threads {threads} -O BAM -f 4 {input.bam} \
+-o {params.tmpdir}/{wildcards.sample_type}.um.bam ;
+
+samtools index {params.tmpdir}/{wildcards.sample_type}.um.bam ;
+
+samtools view --threads {threads} -O BAM -F 4 {params.tmpdir}/{wildcards.sample_type}.um.bam \
+-o {params.tmpdir}/{wildcards.sample_type}.m.bam ;
+
+samtools index {params.tmpdir}/{wildcards.sample_type}.m.bam ;
```
fevac, line 37: can we simplify the `samtools view` commands in a single command?

BALSAMIC/snakemake_rules/align/postprocess_bam.rule
```diff
@@ -0,0 +1,47 @@
...
+
+picard -Xmx75g AddOrReplaceReadGroups -RGPU ILLUMINAi -RGID {wildcards.sample_type} -RGSM {wildcards.sample_type} -RGPL ILLUMINAi -RGLB ILLUMINAi -MAX_RECORDS_IN_RAM 1000000 -CREATE_INDEX true -CREATE_MD5_FILE true \
```
fevac, line 39: I guess we don't know if it's the filtering or the picard command what "makes it work" right?

BALSAMIC/snakemake_rules/align/postprocess_bam.rule
```diff
@@ -0,0 +1,47 @@
...
+-TMP_DIR {params.tmpdir} \
+-INPUT {params.tmpdir}/{wildcards.sample_type}.m.bam  \
+-OUTPUT {output.postprocessed_bam}; 
+
+samtools index {output.postprocessed_bam}; 
+
+rm -rf {params.tmpdir};
```
fevac, line 46: Not a huge fan of these `rm -rf`

BALSAMIC/snakemake_rules/align/postprocess_bam.rule
fevac: or better said, remove the unmapped subset. Is it used?

fevac:
@mathiasbio this is really great. As far as I can see, this PR:
- Increases memory for Vardict
- Filter bam files by only keeping mapped reads
- Changes readgroup in bam files

I guess you don't know what is actually solving the issue?

BALSAMIC/snakemake_rules/align/postprocess_bam.rule
mathiasbio: Good thought! But I think in this particular case they are 2 different ones because 1 is to save only the unmapped reads, and the other is to remove them. So they output to two different bamfiles. I just copied this from the current production, but it can be argued if we should keep these reads or not. I think it's fine to keep them but I don't care that much 🤷

mathiasbio: Honestly...maybe it can be removed entirely. I probably should have explained this as well, that this post-processing is only done for VarDict, the original bamfile will proceed as normal to the other steps. And based on that I think we can just remove this extra samtools command (since the unmapped reads will already be saved in the other bam...)

BALSAMIC/snakemake_rules/align/postprocess_bam.rule
mathiasbio: No, but I have not any evidence at the moment that the removal of unmapped reads helps. So I decided to remove it. We'll see if the next batch of tests works!

BALSAMIC/snakemake_rules/align/postprocess_bam.rule
```diff
@@ -0,0 +1,39 @@
...
+    params:
+        tmpdir = tempfile.mkdtemp(prefix=tmp_dir),
+        sample_id = "{sample}"
+    threads:
+        get_threads(cluster_config, "postprocess_bam")
+    message:
+        "Collapse readgroups for sample: {params.sample_id}"
+    shell:
+        """
+export TMPDIR={params.tmpdir};
```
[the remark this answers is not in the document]
mathiasbio: Hm I'm not sure exactly how that "export TMPDIR" works but we have it in in a lot of different rules (all of the ones I checked where we're using the`tmpdir = tempfile.mkdtemp(prefix=tmp_dir)`), and it seems to work. But it doesn't exactly make sense to me why we need to put it there. I will ask Vadym and Khurram!

BALSAMIC/snakemake_rules/align/postprocess_bam.rule
mathiasbio: Update, removed the part that removes unmapped reads.

BALSAMIC/snakemake_rules/align/postprocess_bam.rule
mathiasbio: Me neither 🤔 I think it has something to do with us using immediate submit instead of sequential submits with snakemake, that the standard use of temp-dirs don't work. Usually they would be automatically cleaned, but I think this might be the only way that they are cleaned. Maybe it's not necessary though!

mathiasbio:
All tests succeed! So I will squash and merge this 🌟

mathiasbio:
And yes! Exactly, I still don't know why it's happening or why this seems to have solved the issue 😬

PR #1334, opened 2023-11-30
fix: remove extra samtools sort from wgs

mathiasbio:
### This PR:

The extra added:
```
samtools sort --threads {threads} -o {output.bam} {input.bam};
samtools index -@ {threads} {output.bam};
```
Was only intended to fix the VarDict bug. It's unclear at the moment if this command added any benefit, and it may slow down a WGS analysis unnecessarily if we add it. 

Added: for new features.
Changed: for changes in existing functionality.
Deprecated: for soon-to-be removed features.
Fixed: for any bug fixes.
Security: in case of vulnerabilities.

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


mathiasbio:
I think that this minor change is safe to merge into the release branch given that the only errors observed were key_errors

Issue #1180, opened 2023-06-16
Remove MAX AF >= 1 filter from bcftools SNV and InDel filters in all workflows


mathiasbio:
## Need

Background to this feature can be found here: https://github.com/Clinical-Genomics/BALSAMIC/issues/1166

In short the need is to remove this filter bcftools filter`--include FORMAT/AF[0] < 1 --soft-filter balsamic_af_one --mode +` which occurs in all workflows for filtering of SNVs and InDels, used here: in [sentieon_quality_filter.rule]( https://github.com/Clinical-Genomics/BALSAMIC/blob/master/BALSAMIC/snakemake_rules/variant_calling/sentieon_quality_filter.rule) 

This should be removed because it's a risk, especially in very pure and somewhat lower-coverage tumor-samples, that a true somatic variant may reach 1 AF and be filtered out.

## Suggested approach

The suggested approach is simply to remove it from the rule. However, we want to make sure first that the filter is not terribly important for filtering out false positives so that we can if necessary make a more sophisticated solution.

## Considered alternatives
*Were there alternative approaches which have been rejected?*

## Requests/suggestions/bugs solved by the feature
- https://github.com/Clinical-Genomics/BALSAMIC/issues/1166

## Can be closed when
*Link the issues needed to be closed for this to be implemented*

## Blockers
*Anything preventing this from happening?*

mathiasbio:
On raw VCFs generated from vardict and TNscope with the most recent version of balsamic 12.0.2 I have run bcftools the same way we do in the TGA and WGS workflow.

For TGA:
`singularity exec varcall_py3.sifbcftools filter --include 'INFO/AF < 1' --soft-filter 'balsamic_af_one' --mode '+' -o ${1}_balsamic_af_one.vcf $1`

For WGS:
`singularity exec varcall_py3.sif bcftools filter --include 'FORMAT/AF[0] < 1' --soft-filter 'balsamic_af_one' --mode '+' -o ${1}_balsamic_af_one.vcf $1`


<google-sheets-html-origin>

analysis_type | filename | total variants | balsamic_af_one | percent balsamic_af_one
-- | -- | -- | -- | --
WES-T | SNV.somatic.case.vardict.vcf.gz | 1416117 | 9435 | 0,67%
WES-T | SNV.somatic.case.vardict.vcf.gz | 125365 | 22618 | 18,04%
[...]

PR #1338, opened 2023-12-05
feat: remove max af 1 from all workflows

mathiasbio:
### This PR:

This change has been made to remove the max allele frequency 1 filter, which has a risk of removing true somatic variants at a very high tumor purity. See issue: https://github.com/Clinical-Genomics/BALSAMIC/issues/1180

- Removed bcftools command for setting soft-filter with maximum allele-frequency, which eventually leads to a hard removal of these variants. 

## Tests:

- [x] All analysis-types completed successfully
- [x] Few extra variants as a result of removing this filter

| sequencing type | T/TN         | case         | # PASS variants release 13 | # PASS variants this PR | # additional PASS variants | % extra PASS variants |
|-----------------|--------------|--------------|----------------------------|-------------------------|----------------------------|-----------------------|
| WGS             | T            | civilsole    |                      33051 |                   33051 |                          0 |                     0 |
| WGS             | T            | firstviper   |                      54072 |                   54072 |                          0 |                     0 |
| WGS             | TN           | fleetjay     |                       7189 |                    7190 |                          1 |                 0,01% |
| TGA             | T            | setamoeba    |                       2415 |                    2416 |                          1 |            0,04%           |
| TGA             | TN           | unitedbeagle |                       1573 |                    1573 |                          0 |                     0 |
| TGA UMI         | T (TNscope)  | equalbug     |                        158 |                     158 |                          0 |                     0 |
| TGA UMI         | T (VarDict)  | equalbug     |                         70 |                      70 |                          0 |                     0 |
| TGA UMI         | TN (TNscope) | uphippo      |                        124 |                     124 |                          0 |                     0 |
| TGA UMI         | TN (VarDict) | uphippo      |                        105 |                     105 |                          0 |                     0 |

Note that likely many variants would have been added in the T-only WGS cases if at the same time the T-only WGS specific filter `bcftools filter --threads {threads} --include 'FORMAT/ALT_F1R2 > {params.strand_reads[0]} && (FORMAT/ALT_F1R2 > 0  && FORMAT/ALT_F2R1 > {params.strand_reads[0]} &&  FORMAT/REF_F1R2 > {params.strand_reads[0]} && FORMAT/REF_F2R1 > {params.strand_reads[0]})' --soft-filter '{params.strand_reads[1]}' --mode '+' 
` was also removed. But as this filter also requires that some reads support a reference variant, then removing the MAX_AF 1 filter in this analysis type has no real effect. 

I think we can conclude with these stats that it is safe to remove the MAX AF filter. 

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


BALSAMIC/models/params.py
```diff
@@ -214,7 +214,6 @@ class VarCallerFilter(BaseModel):
```
ivadym, line 214: Remove it also from the docstring :)

BALSAMIC/snakemake_rules/variant_calling/sentieon_quality_filter.rule
```diff
@@ -94,7 +90,6 @@ if config["analysis"]["sequencing_type"] == 'targeted' and config["analysis"]["a
       AD=[VARDICT.AD.tag_value, VARDICT.AD.filter_name],
       DP=[VARDICT.DP.tag_value, VARDICT.DP.filter_name],
       AF_min=[VARDICT.AF_min.tag_value, VARDICT.AF_min.filter_name],
-      AF_max=[VARDICT.AF_max.tag_value, VARDICT.AF_max.filter_name],
```
ivadym, line 97: We should probably remove it from `VARDICT_SETTINGS` too

ivadym (approved):
Looks good 💯 🎅

BALSAMIC/models/params.py
mathiasbio: Nice catch!

BALSAMIC/snakemake_rules/variant_calling/sentieon_quality_filter.rule
mathiasbio: Ah yes, of course : )

Issue #1299, opened 2023-10-27
Migrate CNV report module from deprecated FPDF2 to pypdf


ivadym:
## Need
The need is to address the deprecation of `fpdf2`, which has moved to the @py-pdf GitHub organization (https://github.com/py-pdf/pypdf). We want to ensure the continued functionality of the affected features.

## Suggested approach
- Migrate our code to use the new `pypdf` package (`generate_cnv_report.py` script)

## Can be closed when
- [x]  The `generate_cnv_report.py` script has been successfully migrated to use pypdf. https://github.com/Clinical-Genomics/BALSAMIC/pull/1346
- [x] Removed PYPDF2 dependencies

PR #1339, opened 2023-12-06
feat: TGA CNV report

ivadym:
### Added:
* CNV report for TGA workflow
* `wkhtmltopdf` to system requirements 

### Review and tests: 
- [x] Tests pass (I will test the rest of the workflows  in the upcoming PR migrating the WGS reports)
  - [x] TGA tumor-only: [CNV.somatic.setamoeba.report.pdf](https://github.com/Clinical-Genomics/BALSAMIC/files/13582783/CNV.somatic.setamoeba.report.pdf)
  - [x] WGS tumor-only
- [ ] Code review

### DIFF
--- a/BALSAMIC/snakemake_rules/scripts/generate_pdf.rule
+++ b/BALSAMIC/snakemake_rules/scripts/generate_pdf.rule
@@ -4,19 +4,17 @@
 rule csv_to_pdf:
     """Converting LOH report CSV files to PDF format."""
     input:
-        csv_path="{csv_path}",
+        csv_path="{pdf_path}".replace(FileType.PDF, FileType.CSV),
     output:
-        pdf_path="{csv_path}".replace(FileType.CSV, FileType.PDF),
+        pdf_path="{pdf_path}",
     wildcard_constraints:
-        csv_path="|".join(config_model.get_purecn_csv_paths(cnv_dir)),
+        pdf_path="|".join(config_model.get_purecn_pdf_paths(cnv_dir)),
     params:
         csv_to_pdf_script=get_script_path("csv_to_pdf.py"),
     threads:
         get_threads(cluster_config=cluster_config, rule_name="csv_to_pdf")
     message:
         "Converting CSV file to PDF {output.pdf_path}"
-    benchmark:
-        Path(benchmark_dir, f"csv_to_pdf_{config['analysis']['case_id']}.tsv").as_posix()
     shell:
         """
 python {params.csv_to_pdf_script} {input.csv_path} {output.pdf_path}
--- a/BALSAMIC/snakemake_rules/scripts/generate_pdf.rule
+++ b/BALSAMIC/snakemake_rules/scripts/generate_pdf.rule
@@ -6,15 +6,13 @@ rule csv_to_pdf:
     input:
         purity_csv=f"{cnv_dir}CNV.somatic.{config['analysis']['case_id']}.purecn.purity.csv",
     output:
-        loh_pdfs=expand(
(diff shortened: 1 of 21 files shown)


khurrammaqbool (approved):
Looks good 🥇 
A general comment: The current report is converted from HTML to PDF and we have a PDF report from merged PDF documents in WGS workflow. It will be nice to unify these reports to follow the same format, structure and template etc with the idea that the report should be developed and interpreted in a unified approach. Please feel free to open an issue if needed.

ivadym:
Thanks for the review @khurrammaqbool! That's a good point, and I'm actually working on that right now. I'll unify the CNV reports for release v13 and remove the FPDF deprecated module (https://github.com/Clinical-Genomics/BALSAMIC/issues/1299)

BALSAMIC/snakemake_rules/scripts/generate_pdf.rule
```diff
@@ -0,0 +1,25 @@
+"""Rules for generating PDF reports."""
```
mathiasbio, line 1: I wonder why these output-files are included as params. Maybe they aren't always produced and that's why they have to be made optional like this. Anyway! We can look into that later if it's necessary : )

mathiasbio (approved):
Beautiful implementation! 🌟 My only questions are related to the results and the customers. I don't know to what degree the customers are aware of these results and how to use them. From the user meeting it seemed like it was a bit unknown to them, but maybe it was just that particular customer that didn't know. But I think it would be great if we could show the relevant customers this report, if it hasn't been done already, to get feedback if there is any. Maybe for instance they would love this report but that it would be more easily used if we highlighted some genes / regions in the table etc. From my perspective right now I imagine it would be a bit cumbersome to scroll through the table unless I was really looking for something particular 🤔 But if we do talk to them to get feedback we could implement that in a separate PR I'm sure!! Looks great :))

BALSAMIC/snakemake_rules/scripts/generate_pdf.rule
ivadym: Yes, very good point, I just also came to that conclusion. We need to use checkpoints for this particular case if possible (https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#data-dependent-conditional-execution).

I'll have a closer look and try to fix it or create some conditional executions in these new rules since they will break the analysis if some files are missing 🙃

mathiasbio: Nice, I didn't even know that possibility existed : )

ivadym: Sleeping on it and talking with Eva a bit made us realise that's checkpoints are incompatible with --immediate-submit. I will just make these files optional and keep the original logic 🙃

ivadym:
> Beautiful implementation! 🌟 My only questions are related to the results and the customers. I don't know to what degree the customers are aware of these results and how to use them. From the user meeting it seemed like it was a bit unknown to them, but maybe it was just that particular customer that didn't know. But I think it would be great if we could show the relevant customers this report, if it hasn't been done already, to get feedback if there is any. Maybe for instance they would love this report but that it would be more easily used if we highlighted some genes / regions in the table etc. From my perspective right now I imagine it would be a bit cumbersome to scroll through the table unless I was really looking for something particular 🤔 But if we do talk to them to get feedback we could implement that in a separate PR I'm sure!! Looks great :))

I believe Khurram was in close contact with some customers to develop the method, so I would assume they expect a CNV report with these characteristics. But yes, we can always refine it :)

PR #1340, opened 2023-12-08
chore: update user_story with SOUP field

pbiology:
### This PR:

Changed: Added a SOUp field to the user-story template

### DIFF
--- a/.github/ISSUE_TEMPLATE/user_story.yml
+++ b/.github/ISSUE_TEMPLATE/user_story.yml
@@ -59,7 +59,7 @@ body:
 - type: textarea
   attributes:
     label: SOUPs
-    description: Does this user story introduce or modify an SOUPs? If so, link to the SOUP(s) here.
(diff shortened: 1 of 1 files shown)


ivadym (approved):
Looks good 👍 

I've also reviewed the rest of the content of the user story and I don't think I'm missing anything.

We could also include a field for documentation update (atlas, order portal, balsamic docs), but I'm not sure if it should be added to the PR template or to the system  requirements field instead

ivadym:
> > Looks good 👍
> > I've also reviewed the rest of the content of the user story and I don't think I'm missing anything.
> > We could also include a field for documentation update (atlas, order portal, balsamic docs), but I'm not sure if it should be added to the PR template or to the system requirements field instead
> 
> Good point!
> 
> Spontaneously I'd say we should add it to the PR template as one of the steps involved in implementing the suggested changes.

Nice! I also think we need to reformat our PR template in any case as we're not making the most of it. I will try to come out with a suggestion later this afternoon

PR #1347, opened 2023-12-14
feat: Store WGS raw plots

ivadym:
### Added:
- Store WGS CNV report plots

Hermes changes: https://github.com/Clinical-Genomics/hermes/pull/86/commits/a55e390e5f6c5d36beed633032416a687acbbd7a

### Review and tests: 
- [ ] Tests pass
- [ ] Code review


BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_normal.rule
```diff
@@ -194,13 +194,13 @@ rule ascat_tumor_normal:
     output:
         final_vcf = vcf_dir + "CNV.somatic." + config["analysis"]["case_id"] + ".raw.ascat.vcf.gz",
         ascat_copynumber = vcf_dir + "CNV.somatic." + config["analysis"]["case_id"] + ".ascat.copynumber.txt.gz",
-        sample_statistics = temp(vcf_dir + "CNV.somatic." + config["analysis"]["case_id"] + ".ascat.samplestatistics.txt"),
```
mathiasbio, line 197: I wonder if this temp() solution is working, I think these files were saved anyway? Regardless it's nice to remove the temp() call, it's just confusing to keep it if it's not doing anything :D

mathiasbio: Or was this temp() was used in the delivery function of balsamic to not add these files to the hk-file? And by removing it we're now adding these to there, and then we need the corresponding change in hermes to start to store them? : )

BALSAMIC/constants/rules.py
```diff
@@ -204,6 +204,7 @@
     "delly_cnv_tumor_only",
     "delly_cnv_tumor_normal",
     "ascat_tumor_normal",
+    "cnvpytor_tumor_only",
```
mathiasbio, line 207: And it makes sense to add the tumor only cnvpytor plots as well 👍

mathiasbio (approved):
Looks good to me! Nice job! ❤️

BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_normal.rule
ivadym: Exactly! The deliver function skips them if they are marked as temp and it doesn't include them in the .hk file. 

Regading hermes, I'm working on it now (I'll also update the validation report so we test that these files are stored)

ivadym: Thanks for the review 🙏

ivadym: Hermes commit: https://github.com/Clinical-Genomics/hermes/pull/86/commits/a55e390e5f6c5d36beed633032416a687acbbd7a

Issue #1353, opened 2024-01-12
Incorrect field name Loqusdb somatic SV annotations


ivadym:
**Describe the bug**

Scout requires the following naming convention for Loqusdb annotations (i.e. `Cancer_Somatic_Obs`):
- https://github.com/Clinical-Genomics/scout/blob/258af81b13b47c64c7c848879401448434600d80/scout/parse/variant/variant.py#L205-L224

This hasnt't been considered when implementing  SV Loqusdb annotations (`somatic_obs`) in https://github.com/Clinical-Genomics/BALSAMIC/pull/1194/files.

**Version (please complete the following information):**
`balsamic --version` 13.0.0 (release)

PR #1354, opened 2024-01-12
fix:  Loqusdb SV annotations

ivadym:
### This PR:

Fixes https://github.com/Clinical-Genomics/BALSAMIC/issues/1353

### Review and tests: 
- [ ] Tests pass
- [ ] Code review

### DIFF
--- a/BALSAMIC/snakemake_rules/annotation/somatic_sv_annotation.rule
+++ b/BALSAMIC/snakemake_rules/annotation/somatic_sv_annotation.rule
@@ -128,7 +128,7 @@ rule svdb_annotate_somatic_obs_somatic_clinical_sv:
         """
 if [[ -f "{params.somatic_sv_observations}" ]]; then
     svdb --query --bnd_distance 10000 --overlap 0.80 \
-    --in_occ Obs --out_occ somatic_obs --in_frq Frq --out_frq somatic_frq \
+    --in_occ Obs --out_occ Cancer_Somatic_Obs --in_frq Frq --out_frq Cancer_Somatic_Frq \
     --db {params.somatic_sv_observations} --query_vcf {input.vcf_sv_clinical_obs} > {params.vcf_somatic_obs}
     bgzip -l 9 -c {params.vcf_somatic_obs} > {output.vcf_sv_clinical};
     rm {params.vcf_somatic_obs};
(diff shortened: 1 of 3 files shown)


BALSAMIC/snakemake_rules/annotation/somatic_sv_annotation.rule
```diff
@@ -94,7 +94,7 @@ rule svdb_annotate_clinical_obs_somatic_clinical_sv:
         """
 if [[ -f "{params.clinical_sv_observations}" ]]; then
     svdb --query --bnd_distance 10000 --overlap 0.80 \
-    --in_occ Obs --out_occ clin_obs --in_frq Frq --out_frq Frq \
+    --in_occ Obs --out_occ Obs --in_frq Frq --out_frq Frq \
```
ivadym, line 97: Scout also supports `clin_obs` per request of KM (https://github.com/Clinical-Genomics/scout/issues/4017), but I'm not sure of why we have different names for SNVs and SVs... Any ideas @mathiasbio?

mathiasbio: Hm, I have no idea why they are different but maybe for the moment it could be safer to just let this be 🤔 and we can think about refactoring it later? Since from what I could tell at least this parameter was correctly loaded into Scout. What do you think?

BALSAMIC/snakemake_rules/annotation/somatic_sv_annotation.rule
```diff
@@ -128,7 +128,7 @@ rule svdb_annotate_somatic_obs_somatic_clinical_sv:
         """
 if [[ -f "{params.somatic_sv_observations}" ]]; then
     svdb --query --bnd_distance 10000 --overlap 0.80 \
-    --in_occ Obs --out_occ somatic_obs --in_frq Frq --out_frq somatic_frq \
+    --in_occ Obs --out_occ Cancer_Somatic_Obs --in_frq Frq --out_frq Cancer_Somatic_Frq \
```
mathiasbio, line 131: 🌟

mathiasbio (approved):
Great job! So quick, and you even fixed the docs. Amazing, thank you! ⭐

BALSAMIC/snakemake_rules/annotation/somatic_sv_annotation.rule
ivadym: "OK, I'll leave it as it is and only change the somatic names. They use two different fields in RD:

```
# SNVs contain INFO field Obs, SVs contain clinical_genomics_loqusObs
```
And although I think `clinical_genomics_loqusObs` is a legacy one, let's play it safe and change what's actually not working

Thanks for the review 🧙‍♂️ ⭐

Issue #1364, opened 2024-01-19
[Assessment] Extra variant calls by Manta after new release v13.0.0


mathiasbio:
### Description

After the changes to the core pipeline in V13.0.0 a substantial increase in Manta variants has been observed.

This can be read up on a bit more in the "Number of variants" section of the Balsamic validation report for V13.0.0. (https://github.com/Clinical-Genomics/validations/blob/main/docs/balsamic/balsamic13.0_validation_report.md)

But in summary the number of SVs for two cases was significantly increased after this update:

civilsole, from 8900 to 10200
setamoeba, from 753 to 1460

This has also been observed in production cases, to an even larger extent. Increasing the priority of this assessment.

### Suggested approach

Run the same sample using different bamfiles, in a step-wise process adding and removing changes between V12 and V13 to isolate which update results in the change. Then identify a way to proceed.

### Conclusion (current)

The correction of the trimming in V13, by first trimming UMIs then quality-trimming, resulted in previously soft-clipped bases being hard-trimmed, which probably resulted in them not being ignored by Manta and consequently lead to the extra variant calls.

To fix this we should likely implement some additional filters! 

### Criteria

The number of SV calls from Manta

mathiasbio:
This issue has popped up in production as well. For instance case X (see[ CG doc](https://docs.google.com/document/d/1LPKxFuH-McRFnJ2B68OnPoaiYWygjgrN91sGzwsC35k/edit))

Which is a TGA case with a total clinical SVs of: 15755 the majority of these come from Manta, and checking the original Manta results in the vcf-dir: SV.somatic.X.manta.vcf.gz

There were 15488 Manta variants after the initial calls.

I have re-run the case with the balsamic v12.0.2 version and instead reached 3771 calls, which is still substantial but significantly different from the v13.0.0 calls.

There are a few steps that differ in the generation of the final bamfiles between the two versions, specifically some post-processing steps which have been removed and which I have tested re-applying ot the v13.0.0 bamfile before restarting Manta.

- I tested removing unmapped reads: 15488 calls
- I tested applying fix mate: 15488 calls
- I tested applying add RG: 15488 calls
- I tested applying all of them, replicating all post-processing steps: 15488 calls

On the other hand, re-running the v12.0.2 generated bamfile with this manual manta-script resulted in 3771 calls. In conclusion at the moment it seems that it has nothing to do with the missing post-processing steps. But seems to be more due to some core changes in the alignment itself...

[...]

Issue #975, opened 2022-06-30
FLT3.ITD300 filtered out in manta result


moahaegglund:
**BALSAMIC version**
Version 8, 9, 10

When looking at HD829 reference samples the FLT3.ITD300 variants is sometimes not detected by vardict but is caught with manta. The problem is that the variant is filtered out from manta due to MaxDepth, "Sample site depth is greater than 3x the median chromosome depth near one or both variant breakends". This variant is of great importance and we need to look into changing this filter or possibly add exception in order to include this variant in the final vcf. 

Cases from the validation where this is an issue:
* proudsquid
* holykid

From step 5 of the validation report of the myeloid panel it is possible to find more cases where this occurs, we could use these cases to find a solution to this.

moahaegglund:
Additional HD829 cases where the variant is filtered out in Manta:
* petspider
* majorsloth
* humaneearwig
* superlouse
* maturecat
* usablegoose
* raremarten

Please see table 5.5 in the GMS myeloid validation report if you want to have even more cases to look at.

PR #1371, opened 2024-01-26
feat: filter manta calls

mathiasbio:
## Description

In here a number of different issues will be solved by a few small changes. 
1. The original issue where the solution of keeping MaxDepth filtered Manta variants only had TGA cases, and the reason they were labeled as MaxDepth is because they weren't run with "--exome" option in Manta as is recommended by Manta docs. By adding this --exome flag we shouldn't be required anymore to have extra bcftools filters to rescue MaxDepth variants.
2. The bcftools filter that aims to keep MaxDepth variants is not correct which results in keeping variants with additional filters as well, by removing the MaxDepth rescue filter entirely, these should be correctly filtered out.
3. Manta variants called on TGA has a lot of variants with MaxDepth with an additional side effect of having PR:SR 0,0:0,0. Here I add --exome tag to Manta rules for TGA to get the correct stats.
4. The majority of Manta calls in some samples have very low evidence, with sometimes more than 90% of variants being below 1% VAF and with very little read support. To solve this a bcftools filter command was added to the Manta rules, requiring at least 4 reads to support the variant (which is now possible to do because MaxDepth isn't added, which preserves the PR:SR info)

Relevant issues:
- Original FLT3 issue: https://github.com/Clinical-Genomics/BALSAMIC/issues/975
- Extra variants from Manta after changing trimming settings in V13 https://github.com/Clinical-Genomics/BALSAMIC/issues/1364
- And the associated user-story for the bcftools filter: https://github.com/Clinical-Genomics/BALSAMIC/issues/1391
- User-story for removing MaxDepth filters and adding --exome flag https://github.com/Clinical-Genomics/BALSAMIC/issues/1392

#### Added

- bcftools filter to all Manta rules to remove all variants with support from fewer than 4 reads.
- "--exome" flag to Manta runs on TGA cases to remove MaxDepth filter

#### Changed

- None

#### Fixed

- None

#### Removed

- MaxDepth from list of filters to keep in bcftools view commands

## Documentation

- [ ] N/A
    - docs/balsamic_sv_cnv.rst (added information about new filters, and removed info about keeping MaxDepth)

#### Verify correct Manta settings and filters

**Manta setting `--exome` should only be present in TGA cases**

See TGA settings in screenshot below:

cleanfowl T only:

image

(exome argument set)

See WGS setting in screenshot below:

Screenshot 2024-02-09 at 08 58 45

(no exome argument set)

**Bcftools filter `low_pr_sr_count` should be correctly applied:**

See screenshot of example header from Manta T+N rule and Manta T-only rules below:

cleanfowl TGA T-only

image

case H WGS T-only

image

See example of applied filter in VCF example variant below in case sweetelf:

**low_pr_sr_count**
`5	84761092	MantaINV:13082:1:2:0:1:0	C	<INV>	.	low_pr_sr_count	END=99239925;SVTYPE=INV;SVLEN=14478833;CIPOS=0,1;CIEND=-1,0;HOMLEN=1;HOMSEQ=A;INV3	PR:SR	774,1:2188,2`

PR:SR = 774,1:2188,2
Alt PR:SR = 1 + 2 = 3 smaller than 4 filter should be set!

**PASS**
`22	37332762	MantaBND:23188:1:2:0:0:0:0	T	T]10:35058943]	.	PASS	SVTYPE=BND;MATEID=MantaBND:23188:1:2:0:0:0:1;CIPOS=0,6;HOMLEN=6;HOMSEQ=CAGGGA;BND_DEPTH=3401;MATE_BND_DEPTH=6	PR:SR	475,1:1146,4`

PR:SR = 475,1:1146,4
Alt PR:SR = 1 + 5 = 6 larger than 3 filter should not be set!

Validation status:

-   [x] Successful

##### Criteria

In all downsampled HD829 cases tested the following criteria is met:

-   [x] FLT3:p.ITD300 is detected with manta in the same cases it was detected previously.
-   [x] FLT3:p.ITD300 is detected in sweetelf

##### Results

First row is validation of V13.0.0
Bottom row is the results from this PR

| Gene:p.aa     | Chr Position                           | Mutation type | cleanfowl           | proudsquid                | modestjaguar        | dearmarmot          | holykid                   | pass/fail |
| ------------- | -------------------------------------- | ------------- | ------------------- | ------------------------- | ------------------- | ------------------- | ------------------------- | --------- |
| FLT3:p.ITD300 | 13:28608030-28608060 (around 28608043) | DUP (~300bp)  | VarDict 297 length (pos: 28608040), Manta 284 length (pos: 28608046) | Manta+DellySV: 284 length | Vardict: 294 length, Manta and DellySV 284 length (pos: 28608046) | Vardict: 292 length, Manta and DellySV length 284 (pos: 28608046) | Manta+DellySV: 284 length | pass      |
| FLT3:p.ITD300 | 13:28608030-28608060 (around 28608043) | DUP (~300bp)  | VarDict 297 length (pos: 28608040), Manta 284 length (pos: 28608046) |  Manta and DellySV 284 length (pos: 28608046) |  VarDict 294 length (pos: 28608042), Manta and DellySV 284 length (pos: 28608046) | VarDict 292 (pos: 28608043), Manta and DellySV length 284 (pos: 28608046)  |  Manta and DellySV 284 length (pos: 28608046)  |  pass     |

As a side-note some of the above variants were not detected in Manta or DellySV previous to version 13.0.0, and were most likely rescued due to the inclusion of more informative reads for SV calling.

Sweetelf comparison of FLT3 detection in v13.0.0 and current PR.

|         | caller  | case     | variant  | chromosome | position | ref | alt       | VAF    |
|---------|---------|----------|----------|------------|----------|-----|-----------|--------|
| v13.0.0 | VarDict | sweetelf | p.ITD300 |         13 | 28608043 | -   | 291bp-ins | 0,0266 |
| v13.0.0 | Manta and DellySV   |  sweetelf | p.ITD300        |    13      |    28608046        |    -      |   284bp-ins  |     0 due to MaxDepth      |       
| this PR | VarDict | sweetelf         |   p.ITD300       |      13      |     28608043     |   -  | 291bp-ins      |    0.0266   |
| this PR | Manta and DellySV | sweetelf   |   p.ITD300        |    13      |    28608046        |    -      |   284bp-ins  | PR:SR 1605,32:5837,137  | 

Validation status:

-   [x] Successful

##### Criteria

The following criteria need to be met:

- [x] For all known variants in the table below, all need to have a PR:SR above 3 in the tumor, and not have MaxDepth filter set in the previous validation. As nothing else has changed in the current PR, verifying this is sufficient to understand that these variants would not be filtered out in the update. 

##### Results

original case-ids hidden, find real case names here: https://docs.google.com/document/d/12x3ozQE62bk_yWNTBW_u8vZMqhpKFRMW3yizzM_mWZs/edit

| Mutation                   | Case        | SVTYPE | Detected ininitally in balsamic version | Detected previously by tool      | v13.0.0  Chr;Pos 1   | v13.0.0  Chr;Pos 2   | v13.0.0 PR:SR values | v13.0.0 filter |  Pass / Fail  |
| -------------------------- | ----------- | ------ | --------------------------------------- | -------------------------------- | ----------- | ----------- | -------------------- | -------------------- | ------------------------------------- | 
| LAMB1-AXIN1 fusion protein | A  | BND    | 10.0.2                                  | manta, dellysv, tiddit           | 7;107609345 | 16;365170   |    145,31:208,29      |     PASS       |    PASS                    |
| NAB2-STAT6 fusion          | B | INV    | 10.0.2                                  | manta, dellysv, tiddit           | 12;57488292 | 12;57493354 |     89,28:118,33    |   PASS      |     PASS                           |
| GAB1-ABL1                  | C | BND    | 11.0.2                                  | manta, dellysv, tiddit           | 4;144373942 | 9;133657503 |     60,38:86,40     |  PASS       |      PASS                   |
| inv(2);RANBP2::ALK         | D  | INV    | 11.0.2                                  | manta, dellysv, tiddit           | 2;102611407 | 2;109375698 |   86,37:116,43      |    PASS      |                   PASS              |
| Del: CDKN2A+CDKN2B         | E | DEL    | 11.0.2                                  | manta, dellysv, tiddit, dellycnv | 9;21878995  | 9;22148438  |  30,33:35,25       |   PASS     |              PASS                    |
| Del: IKZF1 (exon 2-7)      | F  | DEL    | 10.0.2                                  | manta, dellysv, tiddit, dellycnv | 7;50347374  | 7;50463638  |    21,68:23,44       |     PASS   |   PASS                            |

Validation status:

-   [x] Successful

#### Validation of detection of FLT3 in WGS cases

original case-ids hidden, find real case names here: https://docs.google.com/document/d/12x3ozQE62bk_yWNTBW_u8vZMqhpKFRMW3yizzM_mWZs/edit

This has not been previously tested in Balsamic, but WGS cases with known FLT3-ITD has been requested from cust110 and 4 WGS cases with known FLT3-ITD have been shared:  
- G, tumor only - previously found in WGS and in myeloid panel
- H, tumor only - previously found in WGS and in myeloid panel
- I,  tumor only - previously found at 8% VAF in panel, but missed in WGS previously
- J, tumor + normal - previously found at 90% VAF in panel, but missed in WGS previously

Out of these 4 shared cases however, only 1, H had a FLT3-ITD variant called by Manta, which will have to serve to investigate the effect of removing MaxDepth on WGS cases with FLT3-ITD. 

##### Criteria

The following criteria need to be met:
- [x] FLT3-ITD variant in WGS case H is detected in this PR.

##### Results

| observed in balsamic version | case    | found by caller | genomic position | VAF                                   | variant type      |
|------------------------------|---------|-----------------|------------------|---------------------------------------|-------------------|
| 12.0.0                       | H | Manta, TNscope  | 13, 28608221     | Manta PR:SR 0,0:20,37 ; TNscope 0.405 | 69 base insertion |
| this PR                      | H |      Manta, TNscope            |    13, 28608221              |        Manta PR:SR 0,0:20,37 ; TNscope 0.405                               |    69 base insertion               |

Validation status:

-   [x] Successful

### Pipeline Integrity Tests

    - [x] N/A
    - [ ] Verified
- **TGA T/O Workflow**
    - [ ] N/A
    - [x] Verified (in i.e **sweetelf**)
- **TGA T/N Workflow**
    - [ ] N/A
    - [x] Verified (in **K**)
- **UMI T/O Workflow**
    - [ ] N/A
    - [x] Verified (in **uphippo**)
- **UMI T/N Workflow**
    - [ ] N/A
    - [x] Verified (in **equalbug**)
- **WGS T/O Workflow**
    - [ ] N/A
    - [x] Verified (in case **H**)
- **WGS T/N Workflow**
    - [ ] N/A
    - [x] Verified (in i.e case **B**)
- **QC Workflow**
    - [x] N/A
    - [ ] Verified
- **PON Workflow**
    - [x] N/A
    - [ ] Verified

<details>
<summary></summary>

### Documentation

- **Atlas documentation**
    - [x] N/A
    - [ ] Updated: [Link]
    - [x] N/A
    - [ ] Updated: [Link]

### User Changes

- [ ] N/A
- [x] This PR affects the output files or results.
    - [x] User feedback is considered unnecessary because [Justification: All clinically relevant variants for the affected workflow has been verified to be detected, and a minimum read-support to support variant-calling is mostly an issue of quality and a decision that ought to be left for the bioinformatician to make -- unless desires has been made clear by geneticist to receive more variant-calls for SVs of dubious quality, which in this case has been the opposite. Feedback has been provided by geneticists that we are delivering too Many SVs of low quality and that they want help with some pre-filtering which this is doing in a limited way.].

### Infrastructure Changes

    - [x] N/A
    - [ ] Updated: [Link]
    - [x] N/A
    - [ ] Updated: [Link]
    - [x] N/A
    - [ ] Updated: [Link]
- **Scout interface**
    - [x] N/A
    - [ ] Updated: [Link]

### Integration Tests

- [x] N/A
- [ ] Test [Description]
    - [Screenshot]

</details>

## Checklist

> [!IMPORTANT]  

### For Developers

- **PR Description**
- **Documentation**
- **Tests**
    - [ ] Ensured integration of the new code with existing workflows.
- **Review**
    - [x] Addressed and resolved all the feedback provided during the code review process.

### For Reviewers

- **Code**
- **Documentation**
- **Tests**
- **Review**


ivadym:
Let's use develop as the base branch. Maybe we'll include another PR or two for v14 :)

BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_only.rule
```diff
@@ -40,7 +43,9 @@ python {params.tmpdir}/runWorkflow.py -m {params.runmode} -j {threads};
   {input.fa} \
   {params.tmpdir}/results/variants/tumorSV.vcf.gz > {params.tmpdir}/results/variants/tumorSV_converted.vcf; 
 
-bgzip -l 9 -c {params.tmpdir}/results/variants/tumorSV_converted.vcf > {output.final};
+bgzip -l 9 {params.tmpdir}/results/variants/tumorSV_converted.vcf;
+
+bcftools filter --threads {threads} --exclude 'SUM(FORMAT/PR[0:1]+FORMAT/SR[0:1]) < {params.low_pr_sr_count[0]}' --soft-filter '{params.low_pr_sr_count[1]}' --mode '+' -o {output.final} -O z {params.tmpdir}/results/variants/tumorSV_converted.vcf.gz
```
fevac, line 48: should this be a different rule?

tests/models/test_params_models.py
```diff
@@ -18,6 +22,44 @@
...
 
+def test_params_manta():
+    """test Manta settings model for correct validation."""
+
+    # GIVEN Manta params
+    test_manta_params = {"wgs_settings": "", "tga_settings": "--exome"}
+
+    # WHEN building the model
+    test_manta_built = ParamsManta(**test_manta_params)
+
```
fevac, line 33: Add `# THEN` statement

tests/models/test_params_models.py
```diff
@@ -18,6 +22,44 @@
...
+    assert test_manta_built.tga_settings == "--exome"
+    assert test_manta_built.wgs_settings == ""
+
+
+def test_get_manta_settings_tga():
```
fevac, line 38: nice! ⭐

fevac (commented):
Looks good! The only thing is that I think it would be better if the manta filters were in a separate rule and not included in the manta rule. I understand this would mean more work though. Alternatively I wonder if the filtering step can be added to the svdb filtering, although that would potentially filter SVs from other tools too. But would be cleaner in my opinion

BALSAMIC/snakemake_rules/annotation/varcaller_sv_filter.rule
```diff
@@ -23,9 +23,8 @@ rule bcftools_filter_sv_research:
     "Filtering merged research structural and copy number variants using bcftools for {params.case_name}"
   shell:
     """
-bcftools view --threads {threads} -f .,PASS,MaxDepth {input.vcf_sv_research} |\
 bcftools filter --threads {threads} --include 'INFO/SWEGENAF <= {params.swegen_freq[0]} || INFO/SWEGENAF == \".\"' --soft-filter '{params.swegen_freq[1]}' --mode '+' |\
-bcftools view --threads {threads} -f .,PASS,MaxDepth -O z -o {output.vcf_pass_svdb};
+bcftools view --threads {threads} -f PASS -O z -o {output.vcf_pass_svdb};
```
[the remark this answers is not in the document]
mathiasbio: Good question, I actually took a second look while you were reviewing and decided to add it back. It seems that "." is what CNVkit uses instead of PASS to designate it's variants, however, it may be that this is converted at some point to "PASS", maybe during SVDB merge because I have looked at all clinical.filtered.pass vcfs and quite a few other intermediate VCFs after merge, and I haven't seen any variants labeled with ".". 

For instance I did this just now on all clinical.pass.vcfs from the previous validation, and then counted the number of variants in the filtered vcfs which was 0 in all files.

for vcf in $(ls | grep pass.vcf.gz) ; do  singexec /home/proj/stage/cancer/balsamic_cache/12.0.2/containers/varcall_py3.sif bcftools view --threads 2  -f . $vcf > ${vcf}.testout_only_dot.vcf ; done

However, just in case I have added it back, as the "." is sometimes used in callers instead of PASS when there is no filter set. Though I don't think it will have any influence on our results.

BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_only.rule
ivadym: We cannot do the filtering here if we want to maintain the integrity of the raw, research, and clinical VCFs. We'll be adding filtered variants to the raw SVDB file!

tests/models/test_params_models.py
```diff
@@ -18,6 +22,45 @@
 )
 
 
+def test_params_manta():
+    """test Manta settings model for correct validation."""
```
ivadym, line 26:
```suggestion
    """Test Manta settings model for correct validation."""
```

tests/models/test_params_models.py
```diff
@@ -18,6 +22,45 @@
...
+    # WHEN building the model
+    test_manta_built = ParamsManta(**test_manta_params)
+
+    # THEN string values should be correctly populated into the model
+    assert test_manta_built.tga_settings == "--exome"
+    assert test_manta_built.wgs_settings == ""
+
+
+def test_get_manta_settings_tga():
+    """test get Manta settings based on sequencing type TGA."""
```
ivadym, line 40:
```suggestion
    """Test get Manta settings based on sequencing type TGA."""
```

tests/models/test_params_models.py
```diff
@@ -18,6 +22,45 @@
...
+
+    # WHEN getting manta settings for TGA
+    manta_settings = params.get_manta_settings(SequencingType.TARGETED)
+
+    # THEN manta setting should specify exome argument
+    assert manta_settings == "--exome"
+
+
+def test_get_manta_settings_wgs():
+    """test get Manta settings based on sequencing type WGS."""
```
ivadym, line 53:
```suggestion
    """Test get Manta settings based on sequencing type WGS."""
    
```

BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_only.rule
mathiasbio: What will happen here if I understand it correctly in terms of the order:
- Manta calls with these soft filters
- Merged into SVDB raw VCF with all other callers (which will then contain these soft filters)
- Merged SVDB file is filtered to only keep PASS or "." into the SV research VCF (which will remove all variants merged with Manta and which was tagged with the filter)
- SV research VCF is annotated with VEP and gnomad frequencies into the VEP dir
- SV research VCF is filtered into SV.research.filtered vcf
- SV.research.filtered vcf is annotated with clinical database frequencies to the SV.clinical.vcf
- SV.clinical.vcf is filtered to only keep variants outside the population database frequencies, as SV.clinical.filtered_pass.vcf

I suppose the question is:
-  if we want to use the Manta filter to also apply to variants that have been called by other callers and merged due to being of the same type, and being similar in terms of genomic position.
- if we want to apply the Manta filter before merging into SVDB and then only merge Manta variants that are PASS based on this new filter.

Does that sound right @ivadym ? 🌟

mathiasbio: At the moment this Manta filtering approach is the same as what we did for TIDDIT. Where we merge soft-filtered variants into the raw SV.merged vcf, which is then hard-filtered downstream before VEP annotation

mathiasbio: But I don't know what this raw merged VCF is intended to be used for. If it's important that they contain more true somatic variants, maybe we should hard-filter the TIDDIT and these Manta-variants before merging. If the intention is for researchers to use this VCF as some type of maximally unfiltered VCF for some research purpose I can't grasp at the moment, then I suppose adding soft-filtered variants is ok 🤷 But I think someone needs to illuminate me on what this SVDB raw file is used for 😬  🤪

.DS_Store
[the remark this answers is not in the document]
mathiasbio: These sneaky files! 🧙

BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_only.rule
mathiasbio: But I suppose the alternative is the move this bcftools filter to after the SVDB merge to this rule: bcftools_quality_filter_svdb

Where we're currently doing this: 
"Merged SVDB file is filtered to only keep PASS or "." into the SV research VCF"

I was a bit hesitant to do that because the SVDB merge files are a bit messy since we're merging formats from different callers into one, and we were looking for a fix specifically for Manta in this. But I've tested it a bit now and it looks like the bcftools filter is working as expected in ignoring variants without the PR:SR field and only setting the filters on the Manta variants

ivadym: Perfect summary! 

It's a good question: What exactly is our raw SVDB VCF 😆 ?

 According to the description on the order portal:

| File                                          | Description                                                                                            |
|-----------------------------------------------|--------------------------------------------------------------------------------------------------------|
| svdb.vcf.gz                                  | Raw VCF containing the merged output of SV (Manta, Delly) and CNV (CNVkit, AscatNgs, Delly, CNVpytor) callers |
| svdb.research.filtered.pass.vcf.gz           | Research annotated VCF with merged, filtered and passed somatic SVs, CNVs and InDels                   |
| svdb.clinical.filtered.pass.vcf.gz          | Clinical annotated VCF with merged, filtered and passed somatic SVs, CNVs and InDels                   |

If I remember correctly the discussions about these VCFs, the intention here was to collect the raw outputs of the variant callers even though it would imply including variants with lower confidence levels and noise. If you can't gasp the purpose of these files, maybe our customer can't either. Should we discuss this in a refinement?

In any case, this is an example of hard filtering, so I'm not sure if we can leave it like that. We could either switch our strategy regarding these files or perform the filtering on the svdb file, as you pointed out

mathiasbio: I forgot that there's an orderportal with some descriptions. That's nice. 

I guess in the case of the TIDDIT filters we strictly limited it to removing normal variants, which can be argued is then almost keeping the raw output of the TIDDIT caller by doing standard normal filtering. And even then we're still adding these variants to the svdb.vcf.gz (raw vcf) with soft-filters, which is the same as we would be doing here with Manta. We're not hard-filtering them, they're only annotated with the low_sr_pr filter (so they're still in the raw svdb file, just with soft-filters added)

If our customers want to use all variants, they still can. In a way we're only adding our suggestions for what are not high confidence variants to the VCF. Which I suppose we're already doing in the SNV raw VCF with filters like the t_lod_star and MLrejected from Sentieon TNscope etc. So we're already kind of doing exactly this 🤔

mathiasbio: It's only in the step before annotating the research VCF that we're actively removing the soft-filtered variants from both TIDDIT and Manta.

ivadym: Yes, I missed that you added the soft filter tag! Then, it doesn't matter so much where you put it, only important for container atomicity :atom:

BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_normal.rule
```diff
@@ -16,11 +16,13 @@ rule manta_tumor_normal:
...
     params:
         tmpdir = tempfile.mkdtemp(prefix=tmp_dir),
+        settings = params.get_manta_settings(sequencing_type=sequencing_type),
         runmode = "local",
         tumor = config_model.get_sample_name_by_type(SampleType.TUMOR),
         normal = config_model.get_sample_name_by_type(SampleType.NORMAL),
         case_name = case_id,
-        manta_install_path = "/opt/conda/share/manta-1.6.0-2"
+        manta_install_path = "/opt/conda/share/manta-1.6.0-2",
+        low_pr_sr_count = [MANTA_FILTERS.low_pr_sr_count.tag_value,MANTA_FILTERS.low_pr_sr_count.filter_name],
```
ivadym, line 25: Split this into two params with descriptive names so it's easier to follow in the bash script

ivadym (approved):
👍 💯 🎸 👨‍🎤 🔥

BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_normal.rule
mathiasbio: I completely agree, I was always wondering why we were doing it like this and found it confusing to read. Then I blindly did the same 🧟

BALSAMIC/snakemake_rules/variant_calling/somatic_sv_tumor_only.rule
fevac: If we want to keep all variants in the SVDB file (which we want) and filter before annotation (which takes time and it's what is causing issues), to me the most logical step would be adding the soft-filtering for manta variants in the manta step (or create a new step after this if we want to keep containers atomic). But I do think it makes sense to include the soft-filter in the svdb file. Then bcftools filter PASS variants to obtain the research vcf.
 
Overall, I think there are too many different filtering and annotation steps in balsamic which makes it hard to follow, but that's what we have!

Issue #1411, opened 2024-03-06
[User Story] <HELP - missed variant in my analysis>


KickiLagerstedt:
### Need

As a clinical geneticist I would like to find disease causing variants when I perform my analysis!

### Suggested approach

When I perform my filtering I would lika to se variants that are NOT COMMON compared to the human genome (Hg19)

### Considered alternatives

No alternatives what so ever!

### Deviation

https://scout.scilifelab.se/cust002/F0042600-2

### System requirements assessed

- [ ] Yes, I have reviewed the system requirements

### Requirements affected by this story

Do not understand...

### Risk assessment needed

- [X] Needed
- [ ] Not needed

### Risk assessment

?

### Anything else?

I have performed an analysis on affected tissue (no normal tissue) and I perform my analysis (Screening UBA1-gene) i get no variants - but if I look att IGV the variant is there (approx heteroz). X:47058450

We have rerun the sample trough MIP and that pipeline detects the variant. 

We would like to use Balsamic for this analysis since we are looking from mosaic variants in this gene - and with low mosaicism that will be missed using MIP. 

Please advice how to proceed.

mathiasbio:
Hi Kicki! Thanks for bringing this to our attention...we completely agree that this variant should not be filtered out. 

I have looked into it and I think the issue here comes from how balsamic has been built to accommodate such varied types of input data, and in the case of exome-samples it has been treated as standard panel-samples where the coverage is assumed to be higher. But it really should have been taken into account from the start that exome-samples generally have a lot lower coverages than smaller panels, and the filtering-strategy should reflect that. We're going to make a quick fix of this and try to deploy it as quickly as possible so it won't happen again.

But in terms of this particular case, is there anything we can do to help?

KickiLagerstedt:
> Hi Kicki! Thanks for bringing this to our attention...we completely agree that this variant should not be filtered out.
> 
> I have looked into it and I think the issue here comes from how balsamic has been built to accommodate such varied types of input data, and in the case of exome-samples it has been treated as standard panel-samples where the coverage is assumed to be higher. But it really should have been taken into account from the start that exome-samples generally have a lot lower coverages than smaller panels, and the filtering-strategy should reflect that. We're going to make a quick fix of this and try to deploy it as quickly as possible so it won't happen again.
> 
> But in terms of this particular case, is there anything we can do to help?

We rerun it through MIP and the variant was called.
[...]

Issue #1396, opened 2024-02-14
[Assessment] Reanalysis (new downsampling) in BALSAMIC where a variant is missed by VarDict


moahaegglund:
### Description

This is about an analysis of a Horizon Discovery reference sample, HD803.

The sample was downsampled and analysed for the first time during 2022 and analysed with BALSAMIC version 8.2.10. The sample was downsampled again to the same amount of reads and reanalysed in version 13.0.0 (and 13.0.1) of BALSAMIC.
One of the listed variants, NRAS:p.Q61K, with an expected allele frequency of 12.5% is picked up by VarDict in the first analysis (BALSAMIC v. 8.2.10) but missed in the second (BALSAMIC v. 13). The variant is clearly visible in IGVviewer in the second analysis and it's also picked up by DNAscope.

### Suggested approach

Would be interesting if you could look into this, more information to identify the sample is available[ here](https://docs.google.com/spreadsheets/d/12lKCdpH58dLLt2EPBMS-flE3prnjmJKWR-wxInk3zz4/edit#gid=0).

mathiasbio:
I completely forgot about this, I'm looking into it now!

mathiasbio:
I have tried re-running VarDict specifically on the bamfile to try to isolate the issue, and I had a lot of issues making it work outside of balsamic. But I managed to at least make it run using standard VarDict, not the VarDict-java that we're using in the snakemake rule. However the downstream filtering using  `teststrandbias.R` failed with errors:

```
Error in fisher.test(matrix(c(d[i, 10], d[i, 11], d[i, 12], d[i, 13]),  : 
  all entries of 'x' must be nonnegative and finite
Execution halted
```
But I output the intermediate results from VarDict and could at least find the variant there.
`
strikingbeagle	1	1	115256530	115256530	G	T	335	39	65	231	6	33	G/T	0.1164	2;2	43.1	1	35.7	1	60.0	38.000	0.1141	0.0060	0	1.000	1	1.0	38	333	CATGGCACTGTACTCTTCTT	TCCAGCTGTATCCAGTATGT	1:115256320-115256699	SNV	0.667	0`

But I don't know if the issue lies with VarDict-java or with the downstream filtering. Maybe we can check and see if we have updated VarDict between this and the previous version... 🤔

mathiasbio:
[...]

Issue #1344, opened 2023-12-11
[Bug] VarDict produces SVs in the SNV file


dnil:
**Describe the bug**
VarDict SVs once again end up in the SNV VCF files, making them hard to visualise for the user when loaded in Scout.

**To Reproduce**
Load e.g. case `novelbear` (or presumably grep for DEL, DUP, BND etc in recent VarDict SNV VCFs).
Or say 
```
zgrep 25956141 /home/proj/production/housekeeper-bundles/novelbear/2023-12-08/SNV.somatic.novelbear.vardict.clinical.filtered.pass.vcf.gz |grep SVLEN |wc -l
1
```

**Expected behavior**
Short nucleotide variants and structural variants should appear in different VCF files.

**Screenshots**
![Screenshot 2023-12-11 at 16 39 32](https://github.com/Clinical-Genomics/BALSAMIC/assets/758570/7a6da16c-142c-4ef4-a6f0-77f10b0a5ae5)

![Screenshot 2023-12-11 at 16 40 29](https://github.com/Clinical-Genomics/BALSAMIC/assets/758570/0191e41f-337f-42b4-9447-286654592992)

**Version (please complete the following information):**

`12.0.2`

**Additional context**
This has been an issue in the distant past, which I believe was not fully solved together with the VarDict devs, but worked around with filtering by Balsamic on the VarDict VCFs into separate files for display & delivery.

mathiasbio:
Hmm I wonder if the fix was maybe only implemented for TNscope (https://github.com/Clinical-Genomics/BALSAMIC/pull/540/files) even though VarDict was mentioned in the original issue (https://github.com/Clinical-Genomics/BALSAMIC/issues/485), and that possibly since VarDict is only used for TGA it hasn't been as much of an issue as for TNscope which is used for WGS since there was less of a chance that SVs would be called in the smaller panel context. I took a little look at some of the VCFs produced by some TGA cases and WES, and it seems that the VarDict SVs are more common in the WES which makes sense since it's a larger panel. 

I suppose this issue has been around for a while and is probably not super urgent, but I'll bring this up on a refinement session. Possibly we should disable the SV calling from VarDict, or separate it and add to the SVDB merge. But since I have no idea how good VarDict is at calling SVs I'm not ready to say that we should do that yet.

mathiasbio:
Decision in refinement meeting 2024-01-12 to just remove the SV calls from VarDict

mathiasbio:
[...]

PR #1414, opened 2024-03-07
feat: exome option and changes to vardict options

mathiasbio:
## Description

There are a few related changes in this PR, all connected to modifications to the VarDict output.
1. An extra option has been added to balsamic `--exome` in order to allow for different variant-calling filters for exome-samples which usually have much lower coverage than the smaller panels. 
2. bcftools filters for exome samples have been changed from standard targeted analysis, most importantly the minimum depth has been decreased from 100 to 20. https://github.com/Clinical-Genomics/BALSAMIC/issues/1411
3. stringency of bcftools filters for TGA samples have been somewhat reduced, reducing min MQ from 40 to 30.
4.  "-u" flag for VarDict has been removed as it could result in wrongful strandbias https://github.com/Clinical-Genomics/BALSAMIC/issues/1396
5. "-U" flag has been removed VarDict tumor-normal, this option disables SV calling and it has been added back due FLT3 calling which requires SV functionality of VarDict to be enabled

#### Added

* new option for exome samples --exome with modified bcftools filters compared to standard targeted workflow 

#### Changed

* reduced stringency of targeted none-exome bcftools filters for min MQ
* removed -u flag from VarDict T+N and T only rules

#### Fixed

- [Description]

#### Removed

* removed -U flag to VarDict T+N rule to start calling SVs in VarDict (required for FLT3)

## Documentation

- [ ] N/A
- [ ] Updated Balsamic documentation to reflect the changes as needed for this PR.
    - [Document Name]

### Feature Tests

- [ ] N/A
- [ ] Test [Description]
    - [Screenshot]

#### Variant-calling quality testing

Detailed results from testing can be found in this sheet on CG google drive: https://docs.google.com/spreadsheets/d/1-CnG0L2J05Z4Phzdird6JKSXCbG_w-O_xOJfpJNTJ70/edit#gid=0

- [x] changes appear to result in an improved analysis quality, or at least not worse based on looking at true positive calls. **Comment:** See "sensitivity and precision TWISTpancancer" in above sheet, a significant improvement which likely comes predominantly from removing the "-u" flag.
- [x] changes in the number of variants called are not significant, or justified below. **Comment**: There is an increase in number of variants called, in some cases more than others, in particular for the exome cases tested, and for one normal TGA case. In the exome cases this is likely due to the lowering of the depth from 100 to 20, and in the case of the TGA case it's most likely due to the removal of the "-u" flag and the associated low insert size of that sample which probably resulted in a lot of variants being filtered out due to artificial strand-bias. 
- [x] significant changes in the number of variants called is justified because: 
- **TGA smaller panels:** with regards to the TGA samples, the quality of the analysis has been assessed using the TWISTpancancer reference samples and the sensitivity has been much improved to comparably little cost to precision. The only question-mark to me is the effect of lowering the MQ from 40 to 30, which resulted in the detection of 2 more true variants, and 38 more FPs (however, these TWIST samples do not have accurate FP detection...so they might also be true). 
- **TGA exome:** with regards to the exome samples, it seems that we are seeing a more signfiicant increase in the number of variants called based on simply adding the "-u" flag. With an about 300% increase in number of variants even without reducing the min depth to 20. After then reducing the min depth to 20 from 100, we get an about 30% increase in variants, which would make sense proportionally as these exome samples were covered at about 70% above 100X. 

##### Further notes from testing:

**Comparatively larger effect of removal "-u" for exome samples**

The number of variants increased more substantially for the exome samples run in comparison to the other TGA-samples (except 1), even without reducing the DP to 20. This is likely due to the same reason for all these samples, a smaller insert-size which may have pushed more variants to beyond the threshold for strand-bias when using VarDict's "-u" option. 

**Further info regarding lowering MQ40 to MQ30**

I looked at about half of the extra 27 variants called in one of the TGA T+N cases in IGV and the majority of them look perfectly fine, and a few looked perhaps questionable. Screenshots are available in the sheet "MQ30-40"

**Further info regarding keeping SVs in the T+N workflow**

I looked at 3 T+N cases used in this testing, and summarising across all final variant-calls in these cases 6 out of 8000 variants were SVs. 

### Pipeline Integrity Tests

    - [x] N/A
    - [ ] Verified
- **TGA T/O Workflow**
    - [ ] N/A
    - [x] Verified multiple times
- **TGA T/N Workflow**
    - [ ] N/A
    - [x] Verified multiple times
 - **TGA T/O Workflow with --exome**
    - [ ] N/A
    - [x] Verified multiple times
- **TGA T/N Workflow with --exome**
    - [ ] N/A
    - [x] Verified multiple times
- **UMI T/O Workflow**
    - [ ] N/A
    - [x] Verified multiple times
- **UMI T/N Workflow**
    - [ ] N/A
    - [x] Verified multiple times
- **WGS T/O Workflow**
    - [x] N/A (NOTHING HAS CHANGED HERE)
    - [ ] Verified
- **WGS T/N Workflow**
    - [ ] N/A  (NOTHING HAS CHANGED HERE, **VERIFIED ONCE ANYWAY**)
    - [x] Verified once
- **QC Workflow**
    - [x] N/A  (NOTHING HAS CHANGED HERE)
    - [ ] Verified
- **PON Workflow**
    - [x] N/A  (NOTHING HAS CHANGED HERE)
    - [ ] Verified

<details>
<summary></summary>

### Documentation

- **Atlas documentation**
    - [x] N/A
    - [ ] Updated: [Link]
    - [x] N/A
    - [ ] Updated: [Link]

### User Changes

- [ ] N/A
- [x] This PR affects the output files or results.
    - [x] User feedback is considered unnecessary because [Justification below].
    - [x] Affected users have been included in the development process and given a chance to provide feedback. **Comment:** The users have not been consulted after the fix was made, but they were the ones essentially requesting this update as important variants were filtered out. With regards to the details of the filters I judge that we are better suited as the experts on the bioinformatics, and knowing better the limits of our truthsets and the different coverage-depths we have for different applications, etc...

### Infrastructure Changes

    - [x] N/A
    - [ ] Updated: [Link]
    - [x] N/A
    - [ ] Updated: [Link]
    - [x] N/A
    - [ ] Updated: [Link]
- **Scout interface**
    - [x] N/A
    - [ ] Updated: [Link]

</details>

## Checklist

> [!IMPORTANT]  

### For Developers

- **PR Description**
- **Documentation**
- **Tests**
- **Review**

### For Reviewers

- **Code**
- **Documentation**
- **Tests**
- **Review**


BALSAMIC/utils/cli.py
```diff
@@ -487,6 +487,12 @@ def get_analysis_fastq_files_directory(case_dir: str, fastq_path: str) -> str:
         return analysis_fastq_path.as_posix()
     return Path(fastq_path).as_posix()
 
+def validate_exome_option(ctx: click.Context, exome: bool):
```
ivadym, line 490: Nice 🔥

CHANGELOG.rst
```diff
@@ -4,13 +4,17 @@
 Added:
 ^^^^^^
 * high_normal_tumor_af_frac filter in bcftools for TNscope T+N filtering out more than 30% TINC https://github.com/Clinical-Genomics/BALSAMIC/pull/1289
+* new option for exome samples --exome with modified bcftools filters compared to standard targeted workflow
+* added -U flag to VarDict T-only rule to not call SVs
```
ivadym, line 8: Would fit better in `changes`, as you're replacing -u with -U to not call SVs

BALSAMIC/constants/variant_filters.py
```diff
@@ -10,15 +10,15 @@
     "description": "General purpose filters used for filtering any variant caller",
 }
 
-# Configuration of VARDICT settings
-VARDICT_SETTINGS = {
+# Configuration of VARDICT settings for smaller panels
+VARDICT_SETTINGS_PANEL = {
```
ivadym, line 14: Maybe we could avoid repetition here by reusing the common parameters:
```
VARDICT_COMMON_SETTINGS = { ... }

VARDICT_PANEL_SETTINGS = {
    **common_settings,
    ...
}

VARDICT_EXOME_SETTINGS = {
    **common_settings,
    ...
}
```

BALSAMIC/workflows/balsamic.smk
```diff
@@ -115,7 +116,13 @@ else:
 
 # Varcaller filter settings
 COMMON_FILTERS = VarCallerFilter.model_validate(COMMON_SETTINGS)
-VARDICT = VarCallerFilter.model_validate(VARDICT_SETTINGS)
+
+# Set VarDict settings depending on if panel is exome or not
+if config_model.panel and config_model.panel.exome:
+    VARDICT = VarCallerFilter.model_validate(VARDICT_SETTINGS_EXOME)
+else:
+    VARDICT = VarCallerFilter.model_validate(VARDICT_SETTINGS_PANEL)
```
ivadym, line 124:
```suggestion
    VARDICT = VarCallerFilter.model_validate(VARDICT_SETTINGS_PANEL)
    
if config_model.panel.exome:
    VARDICT = VarCallerFilter.model_validate(VARDICT_SETTINGS_EXOME)

```

ivadym (commented):
⚡ 🔥

BALSAMIC/workflows/balsamic.smk
mathiasbio: Won't that crash for WGS where the panel model hasn't been created?

BALSAMIC/constants/variant_filters.py
mathiasbio: Done!

CHANGELOG.rst
mathiasbio: But they're two quite different changes, one is to remove the risk of wrongly filtering out variants due to strand-bias, and the other is to not call structural variants, feels like it could be confusing to put them under the same change.

mathiasbio: but it can be quite cluttered if we basically try to document every little change we make in the changelog, so maybe it's better to have some more general summary like "changed vardict settings"

ivadym: Maybe then just clarify the removal of the -u flag. I think it's nice if we're specific, it helps to write the message to the customers and to follow the pipeline development

BALSAMIC/workflows/balsamic.smk
ivadym: Isn't this the current behaviour? We're defining `VARDICT_SETTING` but not using it. It does not rely on the panel model

mathiasbio: The if config_model.panel.exome part is new, and we didn't have that if-statement before we just validated the vardict settings :D But if the panel model hasn't been populated doesn't looking inside it cause some attribute error?

ivadym: Oh you mean the if statement? Yes it should be `if config_model.panel and config_model.panel.exome:`. I just meant to say that the default assignation should be defined outside the `if` so we avoid having an else

mathiasbio: Oh god, of course. That makes sense...I'll make the adjustment ❤️ Thanks!

docs/balsamic_filters.rst
```diff
@@ -109,15 +104,21 @@ Somatic Callers for reporting SNVs/INDELS
 The results of `Vardict` variant calling are further post-filtered based on several criteria (`Vardict_filtering`_) to retrieve high-confidence variant calls.
 These high-confidence variant calls are the final list of variants uploaded to Scout or available in the delivered VCF file in Caesar.
 
+
+There are two slightly different post-processing filters used depending on if the flag "--exome" is supplied at the start, which is designed to accomodate targeted analyses of lower depths.
```
ivadym, line 108: Do we need to be so technical here (im refering to the --exome flag)? Maybe it's clearer if we refer to it as exome samples in general, what do you think?

BALSAMIC/snakemake_rules/variant_calling/somatic_tumor_normal.rule
```diff
@@ -32,7 +32,7 @@ mkdir -p {params.tmpdir};
 export TMPDIR={params.tmpdir};
 export VAR_DICT_OPTS='\"-Djava.io.tmpdir={params.tmpdir}\" \"-Xmx90G\"';
 
-vardict-java -U -u -I 600 -G {input.fa} -f {params.af} -N {params.case_name} \
+vardict-java -I 600 -G {input.fa} -f {params.af} -N {params.case_name} \
```
ivadym, line 35: I'm not sure about removing the -U flag, specially if we are planning to make a release after this PR.

If we want Vardict to call SVs for T/N, we should have an additional file and fed it into SVDB. Same thing for T/O.

ivadym: Then we could also address this issue: https://github.com/Clinical-Genomics/BALSAMIC/issues/1344

But I'm unsure about the effort though

ivadym (commented):
Impressive amount of work! Thanks for documenting everything ⭐

BALSAMIC/constants/variant_filters.py
```diff
@@ -35,12 +27,36 @@
...
     "filter_type": "general",
-    "analysis_type": "tumor_only",
+    "analysis_type": "tumor_only,tumor_normal",
     "description": "General purpose filters used for filtering VarDict",
 }
 
+# Configuration of VARDICT settings for smaller panels
+VARDICT_SETTINGS_PANEL = {
+    **VARDICT_SETTINGS_COMMON,
+    "AD": {"tag_value": 5, "filter_name": "balsamic_low_tumor_ad", "field": "INFO"},
```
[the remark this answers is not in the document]
mathiasbio: Not really! I could move it, in some ways I just liked the idea of having the parameters already separated to encourage creating more specific parameters for the two versions. But it's quite easy to just move them whenever we feel the need as well, so I'll just move it up :  )

BALSAMIC/snakemake_rules/variant_calling/somatic_tumor_normal.rule
mathiasbio: I also don't think it's going to be such a big mess, since it was only quite recently that it was brought up as a bug that we saw SVs in VarDict in Scout, when it had been around for around 3 years already

BALSAMIC/utils/cli.py
```diff
@@ -487,6 +487,12 @@ def get_analysis_fastq_files_directory(case_dir: str, fastq_path: str) -> str:
         return analysis_fastq_path.as_posix()
     return Path(fastq_path).as_posix()
 
+def validate_exome_option(ctx: click.Context, _param: click.Parameter, exome: bool):
```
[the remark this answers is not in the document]
mathiasbio: Done!

tests/utils/test_utils.py
```diff
@@ -1001,6 +1002,28 @@ def test_get_fastp_parameters(balsamic_model: ConfigModel):
...
+    # THEN exome argument should be correctly set
+    assert validate_exome_option(ctx, None, True) == True
+
+    # GIVEN that a panel bedfile has NOT been supplied and exome parameter set to true
+    ctx = MockClickContext({"panel_bed": None})
+
+    # WHEN validating exome option
+    # THEN a bad parameter error should be raised
+    with pytest.raises(click.BadParameter):
+        validate_exome_option(ctx, None, True)
```
ivadym, line 1024: That's a creative way of doing this but I have Sonarlint complaining here: `validate_exome_option(ctx, None, True)`

I think this an be simplified by just using click.Context and click.Patameters. Same as we doing for `test_validate_cache_version_develop` for example.

ivadym, line 1024: Let's use named arguments for function with > 1 parameter

BALSAMIC/snakemake_rules/variant_calling/somatic_tumor_normal.rule
ivadym: Yes, that's true. However, because of the same issue, customers may unexpectedly find SVs in their SNVs tab in Scout for tumor-normal cases.

I definitely don't think we should address this in the current pull request. But if we plan to release after merging this one, our options could be:

- Assess the impact of having these SVs in tumor-normal cases and gather customer feedback.
- Acknowledge it as a known issue and plan to fix it later by separating SVs and SNVs or exploring alternative methods to vardict for SV detection.
- Consider including this user story for the next release: https://github.com/Clinical-Genomics/BALSAMIC/issues/1344

tests/utils/test_utils.py
mathiasbio: I think I figured it out, thanks! :D

BALSAMIC/snakemake_rules/variant_calling/somatic_tumor_normal.rule
mathiasbio: I agree! It would definitely be a lot more cleaner if we could remove the SVs from this file entirely, we would just need to talk to the customers and make sure that they are aware. Shouldn't be that difficult! : )

mathiasbio: I put it in milestone for release 16

ivadym: Did you also check how many extra variants we get in the TGA tumor-normal cases just by keeping the SVs?

Might be worth adding that info to the test section in this PR 🤔

mathiasbio: I didn't specifically look for extra variants from that flag, but I think they're in the scale of 0-10 but I'll verify!

ivadym (approved):
Great work! 💯 🕵️ 🥇

Please update the PR with the number of SVs we're including, so we can keep track of it. Nothing more from my side 😄

BALSAMIC/snakemake_rules/variant_calling/somatic_tumor_normal.rule
mathiasbio: Seems that about 6 variants in 8000 are extra SVs

ivadym: Thanks mathias!

PR #1461, opened 2024-07-02
feat: add Sentieon path argument to config

mathiasbio:
## Description

Adds an argument to balsamic case config which allows for overriding the SENTIEON_INSTALL_DIR in the environment to a chosen path. This would allow us to easier test new versions of Sentieon without needing to make changes in the servers repository, or manually override the environment variables by adding hardcoded paths in the balsamic.smk of the dev-branches.

Together with associated changes in CG we can also start using this argument in production by adding the Sentieon install path to the cg.yaml and supplying it via CG to the case config.

#### Added

- Sentieon install directory path and Sentieon license to case config arguments

## Documentation

- [ ] N/A
    - updated install.rst with new Sentieon argument info

### Feature Tests

- [ ] N/A
- [x] Test that using --sentieon-install-dir and --sentieon-license argument puts these variables in config.json and is working
    - In config.json:

image

   - In alignment rule using the supplied version of Sentieon:
   
image

- [x] Verify that supplying wrong Sentieon license makes the Sentieon rules fail with invalid license

image

### Pipeline Integrity Tests

**Justification for limited pipeline integrity test:**

    - [x] N/A
    - [ ] Verified
- **TGA T/O Workflow**
    - [ ] N/A
    - [x] Verified
- **TGA T/N Workflow**
    - [x] N/A
    - [ ] Verified
- **UMI T/O Workflow**
    - [x] N/A
    - [ ] Verified
- **UMI T/N Workflow**
    - [ ] N/A
    - [x] Verified
- **WGS T/O Workflow**
    - [x] N/A
    - [ ] Verified
- **WGS T/N Workflow**
    - [ ] N/A
    - [x] Verified
- **QC Workflow**
    - [ ] N/A
    - [x] Verified
- **PON Workflow**
    - [ ] N/A
    - [x] Verified

<details>
<summary></summary>

### Documentation

- **Atlas documentation**
    - [x] N/A
    - [ ] Updated: [Link]
    - [x] N/A
    - [ ] Updated: [Link]

### User Changes

- [x] N/A

### Infrastructure Changes

    - [x] N/A
    - [ ] Updated: [Link]
    - [x] N/A
    - [ ] Updated: [Link]
    - [x] N/A
    - [ ] Updated: [Link]
- **Scout interface**
    - [x] N/A
    - [ ] Updated: [Link]

</details>

## Checklist

> [!IMPORTANT]  

### For Developers

- **PR Description**
- **Documentation**
- **Tests**
- **Review**
    - [x] Addressed and resolved all the feedback provided during the code review process.
    - [x] Obtained final approval from designated reviewers.

### For Reviewers

- **Code**
- **Documentation**
- **Tests**
- **Review**


BALSAMIC/commands/config/case.py
```diff
@@ -117,6 +119,7 @@ def case_config(
     panel_bed: Path,
     pon_cnn: Path,
     quality_trim: bool,
+    sentieon_install_dir: Path,
```
ivadym, line 122: Perhaps it makes sense to include both to standardise the implementation. What do you think?

```suggestion
    sentieon_licence_dir: Path,
    sentieon_licence_server: str,
```

mathiasbio: Yeah I don't see why not 🤔 I can't think of a use-case for it at the moment but it could be useful, and it feels good to set these Sentieon variables in a consistent way

BALSAMIC/commands/options.py
```diff
@@ -370,6 +370,20 @@
     help="Sample configuration file",
 )
 
+OPTION_SENTIEON_INSTALL_DIR = click.option(
+    "--sentieon-install-dir",
+    type=click.Path(exists=True, resolve_path=True),
+    required=False,
```
ivadym, line 376: I guess it should be a mandatory field since it's required for our workflows

BALSAMIC/commands/config/case.py
```diff
@@ -188,6 +194,8 @@ def case_config(
         directory.mkdir(exist_ok=True)
 
     config_collection_dict = ConfigModel(
+        sentieon_install_dir=sentieon_install_dir if sentieon_install_dir else None,
+        sentieon_license=sentieon_license if sentieon_license else None,
```
ivadym, line 198: Here as well, it's a mandatory field for balsamic right now. We've been ignoring the initial idea of granularity with the `--disable-variant-callers`

```suggestion
        sentieon_install_dir=sentieon_install_dir,
        sentieon_license=sentieon_license,
```

BALSAMIC/models/config.py
```diff
@@ -194,6 +194,8 @@ class ConfigModel(BaseModel):
         background_variants: Field(Path(optional)); path to BACKGROUND VARIANTS for UMI
         analysis: Field(AnalysisModel); Pydantic model containing workflow variables
         custom_filters: Field(CustomFilters); custom parameters for variant filtering
+        sentieon_install_dir: Field(optional); path to Sentieon installation directory
```
ivadym, line 197:
```suggestion
        sentieon_install_dir: Field(required); path to Sentieon installation directory
```

BALSAMIC/models/config.py
```diff
@@ -194,6 +194,8 @@ class ConfigModel(BaseModel):
...
+        sentieon_license: Field(optional); Sentieon license string
```
ivadym, line 198:
```suggestion
        sentieon_license: Field(required); Sentieon license string
```

BALSAMIC/models/config.py
```diff
@@ -220,6 +222,8 @@ class ConfigModel(BaseModel):
     background_variants: Optional[str] = None
     analysis: AnalysisModel
     custom_filters: CustomFilters | None = None
+    sentieon_install_dir: Annotated[Optional[str], AfterValidator(is_dir)] = None
+    sentieon_license: Optional[str] = None
```
ivadym, line 226:
```suggestion
    sentieon_install_dir: Annotated[str, AfterValidator(is_dir)]
    sentieon_license: str
```

docs/install.rst
```diff
@@ -10,16 +10,18 @@ Software Requirements
...
 
 - Conda >=version 4.5.0: For detailed software and python requirements please see ``setup.py`` and ``BALSAMIC/conda/balsamic.yaml``
-- Singularity >=version 3.0.0: BALSAMIC uses singularity to run vairous parts of the workflow.
+- Singularity >=version 3.0.0: BALSAMIC uses singularity to run various parts of the workflow.
 - Python 3.11
-- BALSAMIC is dependent on third-party bioinformatics software ``Sentieon-tools``. Example: for running wgs variant calling using ``TNScope``, and to execute ``UMIworkflow``.
+- BALSAMIC is dependent on third-party bioinformatics software ``Sentieon-tools`` for all workflows.
 
-``Note: Set Sentieon envionment variables in your ~/.bashrc file by adding following two lines``
+``Note: To run Balsamic you need to either supply the --sentieon-install-dir and --sentieon-license arguments during the config step or set the environment variables in your ~/.bashrc``
```
ivadym, line 17: I'm not sure we should keep the option of defining the license with ENV variables. What's your take on this?

ivadym (commented):
Looks good overall 💯 

I think the review comments may be a bit disorganised, but the main point is to decide if we want to keep the support for the environment variables for the Sentieon license. I don't see any advantages in maintaining that code over time, so I would just stick with the values provided by the CLI

docs/install.rst
mathiasbio: I don't have many opinions on this, I just didn't want to rock the boat too much by default so people could keep testing like usual after the merge into develop, not needing to add extra arguments to their test commands. But I suppose it wouldn't be that much of an effort to add a couple of arguments either. 

Maybe it's good to be more explicit so that we aren't surprised if someone accidentally forgets to supply an argument and it defaults to some old env variable

ivadym: I think I would prefer to make it explicit through the CLI and avoid any unexpected behaviour with double definitions

ivadym: Any preference, @khurrammaqbool?

mathiasbio: I have made it mandatory now and removed the env functionality. There was a lot of things to update in the pytests but they're all passing now 😅

BALSAMIC/commands/options.py
```diff
@@ -370,6 +370,20 @@
...
+    required=True,
+    help="Path to Sentieon install directory",
+)
+
+OPTION_SENTIEON_LICENSE = click.option(
+    "--sentieon-license",
+    required=True,
+    type=click.STRING,
+    help="Sentieon license in format IP:Port",
+)
```
ivadym, line 385: Maybe we could add a more explanatory names for these options:

```suggestion
OPTION_SENTIEON_INSTALL_DIR = click.option(
    "--sentieon-license-path",
    type=click.Path(exists=True, resolve_path=True),
    required=True,
    help="Path to Sentieon install directory",
)

OPTION_SENTIEON_LICENSE = click.option(
    "--sentieon-license-server",
    required=True,
    type=click.STRING,
    help="Sentieon license in format IP:Port",
)
```

BALSAMIC/workflows/PON.smk
```diff
@@ -60,20 +60,10 @@ sample_names: List[str] = config_model.get_all_sample_names()
...
-        config["SENTIEON_EXEC"] = Path(os.environ["SENTIEON_INSTALL_DIR"], "bin", "sentieon").as_posix()
-
-except KeyError as error:
-    LOG.error("Set environment variables SENTIEON_LICENSE, SENTIEON_INSTALL_DIR, SENTIEON_EXEC "
-              "to run SENTIEON variant callers")
-    raise BalsamicError
+# Set Sentieon license and Sentieon install dir environment variables
+config["SENTIEON_LICENSE"] = config_model.sentieon_license
+config["SENTIEON_INSTALL_DIR"] = Path(config_model.sentieon_install_dir).as_posix()
+config["SENTIEON_EXEC"] = Path(config_model.sentieon_install_dir, "bin", "sentieon").as_posix()
```
ivadym, line 66: This should be asigned during the config case. Maybe we could create a model called Sentieon and append all of these attributes there. There is no need to redefine these constants again and again, let's use them directly :)
```suggestion
```

BALSAMIC/workflows/QC.smk
```diff
@@ -78,20 +78,16 @@ if config["analysis"]["sequencing_type"] != "wgs":
...
     cluster_config = config
 
-# Find and set Sentieon binary and license server from env variables
-try:
-    config["SENTIEON_LICENSE"] = os.environ["SENTIEON_LICENSE"]
-    config["SENTIEON_INSTALL_DIR"] = os.environ["SENTIEON_INSTALL_DIR"]
+# Set Sentieon license and Sentieon install dir environment variables
+config["SENTIEON_LICENSE"] = config_model.sentieon_license
+config["SENTIEON_INSTALL_DIR"] = Path(config_model.sentieon_install_dir).as_posix()
+config["SENTIEON_EXEC"] = Path(config_model.sentieon_install_dir, "bin", "sentieon").as_posix()
```
ivadym, line 84: same here

BALSAMIC/workflows/QC.smk
```diff
@@ -78,20 +78,16 @@ if config["analysis"]["sequencing_type"] != "wgs":
...
 
-    if os.getenv("SENTIEON_EXEC") is not None:
-        config["SENTIEON_EXEC"] = os.environ["SENTIEON_EXEC"]
-    else:
-        config["SENTIEON_EXEC"] = Path(os.environ["SENTIEON_INSTALL_DIR"], "bin", "sentieon").as_posix()
-except KeyError as error:
-    LOG.error("Set environment variables SENTIEON_LICENSE, SENTIEON_INSTALL_DIR, SENTIEON_EXEC "
-              "to run SENTIEON variant callers")
+if not Path(config["SENTIEON_EXEC"]).exists():
+    LOG.error("Sentieon executable not found {}".format(Path(config["SENTIEON_EXEC"]).as_posix()))
```
ivadym, line 87: This is already being checked in the pydantic config model

```suggestion
```

BALSAMIC/workflows/balsamic.smk
```diff
@@ -148,6 +148,21 @@ params = BalsamicWorkflowConfig.model_validate(WORKFLOW_PARAMS)
...
+config["SENTIEON_DNASCOPE"] = SENTIEON_DNASCOPE_DIR.as_posix()
+
+if not Path(config["SENTIEON_EXEC"]).exists():
+    LOG.error(
+        "Sentieon executable not found {}".format(
+            Path(config["SENTIEON_EXEC"]).as_posix()
+        )
+    )
+    raise BalsamicError
+
```
ivadym, line 165: same here

BALSAMIC/workflows/PON.smk
ivadym: It also means changing the sentieon params within rules, but should be a quick renaming

docs/install.rst
ivadym: greta work 🙏 🥇

BALSAMIC/commands/options.py
mathiasbio: I agree with the second change of sentieon-license to sentieon-license-server but the sentieon-install-dir I think fits the way we name this variable elsewhere, and I think it's more correct since it's not a license-path but the path to where Sentieon is installed. I could see an argument for --sentieon-install-path but I think the "-dir" part is more in line with how we generally name these arguments in balsamic, with for instance --analysis-dir

mathiasbio: There's also however so many places in the code where sentieon-license would need to be renamed to sentieon-license-server 😓 I don't know if it's worth it

mathiasbio: we're also calling it sentieon_license in the pytests

ivadym: 👍  I don't have a strong opinion on this, it's up to you

ivadym: But I think we'll have to do some refactoring in any case, It would be great to use these options directly from the config and not redefining them

BALSAMIC/workflows/QC.smk
mathiasbio: I don't think it is, it's only checking if the sentieon_install_dir exists but the actual exec is in a subdirectory in sentieon_install_dir/bin/sentieon but I know what you're going to say, we can add this to the sentieon model too 😂

ivadym: Yes 😆 I think that's the proper way of doing it, checking it within the config model. Sorry for being so annoying 🙏

mathiasbio: You're doing your job as a fantastic reviewer, don't apologise: D ❤️

BALSAMIC/commands/options.py
mathiasbio: I have done the refactoring now, replacing the sentieon_install_dir, sentieon_exec, sentieon_license, sentieon_tnscope_dir, sentieon_dnascope_dir from the config to being retreived from a sentieon model ⭐ good suggestion!

mathiasbio: though I didn't change the sentieon_license name, my find and replace in pycharm is being very spotty so it doesn't find all cases of the variables when I search, it doesn't feel worth the time to change. But it still feels like a significant improvement regardless!

BALSAMIC/commands/config/case.py
```diff
@@ -188,6 +198,13 @@ def case_config(
         directory.mkdir(exist_ok=True)
 
     config_collection_dict = ConfigModel(
+        sentieon={
+            "sentieon_install_dir": sentieon_install_dir,
+            "sentieon_license": sentieon_license,
+            "sentieon_exec": Path(sentieon_install_dir, "bin", "sentieon").as_posix(),
+            "dnascope_model": SENTIEON_DNASCOPE_MODEL.as_posix(),
+            "tnscope_model": SENTIEON_TNSCOPE_MODEL.as_posix(),
+        },
```
ivadym, line 207: Beautiful 🌹

BALSAMIC/models/config.py
```diff
@@ -194,6 +211,8 @@ class ConfigModel(BaseModel):
         background_variants: Field(Path(optional)); path to BACKGROUND VARIANTS for UMI
         analysis: Field(AnalysisModel); Pydantic model containing workflow variables
         custom_filters: Field(CustomFilters); custom parameters for variant filtering
+        sentieon_install_dir: Field(required); path to Sentieon installation directory
+        sentieon_license: Field(required); Sentieon license string
```
ivadym, line 215:
```suggestion
        sentieon: Field(Sentieon); Sentieon model attributes
```

ivadym (approved):
Nice! Such a clean implementation at the! 🥇 💯

BALSAMIC/commands/options.py
ivadym: fair enough, I think this is quiet an improvement already from what we had 💯

ivadym:
Tested with the respective CG branch https://github.com/Clinical-Genomics/cg/pull/3388:

Screenshot 2024-07-03 at 14 07 18

PR #1499, opened 2024-11-12
fix: merge snv variants script

mathiasbio:
## Description

**MERGED THESE PRs INTO THIS ONE** to test all new features together:
1. disable normal hardfilter: https://github.com/Clinical-Genomics/BALSAMIC/pull/1509
2. create tnscope mvns: https://github.com/Clinical-Genomics/BALSAMIC/pull/1524

This method is replacing the current bcftools concat method.

The bcftools concat method has a couple of issues:
1. The most important is probably that it doesn't merge INFO fields when the variants are shared in the two VCFs, with the most significant consequence being that we don't keep the info that the variant was called by 2 callers. https://github.com/Clinical-Genomics/BALSAMIC/issues/1518
3. It doesn't require that the variants are matching perfectly in the ALT column, so for instance if a variant has been called as a MNV in VarDict and as separate SNVs in TNscope, it merges only the first variant. https://github.com/Clinical-Genomics/BALSAMIC/issues/1519

Also added this to header of merged vcf:

```
##merge_snv_variantcallers=merge_snv_variantcallers.py SNV.somatic.setamoeba.tnscope.research.normalised.vcf.gz SNV.somatic.setamoeba.vardict.research.normalised.vcf.gz --output output_merged.vcf
##merge_snv_variantcallers_processing_time=2025-01-20T11:18:24
##INFO_MERGE_SNV_VARIANTCALLERS=Values in merged INFO fields are listed in the order of the input files: first from SNV.somatic.setamoeba.tnscope.research.normalised.vcf.gz, then from SNV.somatic.setamoeba.vardict.research.normalised.vcf.gz
```

Also removed the filepath in the FOUND_IN pre-processing by `edit_vcf_info.py`

Based on @khurrammaqbool suggestion I also maintained single values for the AF and DP fields from the 1st VCF in the INFO field and added a new list of AF and DP values which contains the values from both, instead of as previously transforming the AF and DP field directly into a list. 

#### Changed

- Replaced bcftools concat with custom python script for merging VCFs from VarDict and TNscope

## Documentation

- [ ] N/A
    - [balsamic_filters.rst]
    - [balsamic_methods.rst]

### Feature Tests

- [x] Verify that merging works as intended. Will collect test results in this google sheet: https://docs.google.com/spreadsheets/d/1kB2vNaEBmol0tX3HUR3UY1LPQCtWZLutixGwnvbkmhY/edit?gid=0#gid=0
 
Verify that both TNscope and VarDict shows up as callers for merged variants in Scout
- [x] See link: https://scout-stage.scilifelab.se/cust000/10-2-TWISTpancan4
- [x] All variants are uploaded 6969 / 6969

image

Verify that INFO field from VarDict and TNscope variants keep AF and DP as single-value fields for merged variants, and that separate AF and DP LIST fields are created with values from each caller. See sheet: https://docs.google.com/spreadsheets/d/1kB2vNaEBmol0tX3HUR3UY1LPQCtWZLutixGwnvbkmhY/edit?gid=1578036768#gid=1578036768

- [x] Successful

Verify that running the https://github.com/EBIvariation/vcf-validator does not show any new errors that weren't present in the original merged VCF from 16.0.0.

Errors in clinical.filtered.pass vcf from uphippo v16.0.0:

```
According to the VCF specification, the input file is not valid
Error: Error in meta-data section. This occurs 1 time(s), first time in line 265.
Error: Format is not a colon-separated list of alphanumeric strings. This occurs 113 time(s), first time in line 279.
Warning: Reference and alternate alleles do not share the first nucleotide. This occurs 2 time(s), first time in line 442.
```

Errors in clinical.filtered.pass vcf from uphippo this PR:

```
According to the VCF specification, the input file is not valid
Error: Error in meta-data section. This occurs 1 time(s), first time in line 280.
Error: Format is not a colon-separated list of alphanumeric strings. This occurs 115 time(s), first time in line 293.
Warning: Reference and alternate alleles do not share the first nucleotide. This occurs 2 time(s), first time in line 460.
```
- [x] Successful. The "Format is not a colon-separated list of alphanumeric strings" Error is from TNscope and represents the majority of TNscope variants. If this was an issue for uploading to Scout we would have seen it. 

### Pipeline Integrity Tests

    - [x] N/A
    - [ ] Verified
- **TGA T/O Workflow**
    - [ ] N/A
    - [x] Verified
- **TGA T/N Workflow**
    - [ ] N/A
    - [x] Verified
- **UMI T/O Workflow**
    - [ ] N/A
    - [x] Verified
- **UMI T/N Workflow**
    - [ ] N/A
    - [x] Verified
- **WGS T/O Workflow**
    - [x] N/A
    - [ ] Verified
- **WGS T/N Workflow**
    - [ ] N/A
    - [x] Verified
- **QC Workflow**
    - [x] N/A
    - [ ] Verified
- **PON Workflow**
    - [x] N/A
    - [ ] Verified

<details>
<summary></summary>

### Documentation

- **Atlas documentation**
    - [x] N/A
    - [ ] Updated: [Link]
    - [x] N/A
    - [ ] Updated: [Link]

### User Changes

- [x] N/A

### Infrastructure Changes

    - [x] N/A
    - [ ] Updated: [Link]
    - [x] N/A
    - [ ] Updated: [Link]
    - [x] N/A
    - [ ] Updated: [Link]
- **Scout interface**
    - [x] N/A
    - [ ] Updated: [Link]

</details>

## Checklist

> [!IMPORTANT]  

### For Developers

- **PR Description**
- **Documentation**
- **Tests**
- **Review**

### For Reviewers

- **Code**
    - [ ] Code implements the intended features or fixes the reported issue.
    - [ ] Code follows the project's coding standards and style guide.
- **Documentation**
    - [ ] Pipeline changes are well-documented in the CHANGELOG and relevant documentation.
- **Tests**
    - [ ] The author provided a description of their manual testing, including consideration of edge cases and boundary
- **Review**
    - [ ] Confirmed that the developer has addressed all the comments during the code review.


BALSAMIC/assets/scripts/merge_snv_variantcallers.py
```diff
@@ -0,0 +1,486 @@
...
+        updated_line = update_number_text(updated_line, ".")
+        print(
+            f"Warning: Number fields differ. line1: {line1}, line2: {line2}",
+            file=sys.stderr,
+        )
+
+    return updated_line
+
+
+def read_variants(vcf: str) -> Dict[Tuple[str, str, str, str], Dict[str, str]]:
```
fevac, line 226: is this parsing of variants slow and memory consuming?

fevac (commented):
Code looks good to me. I do wonder if this new vcf is compliant with the vcf file format and can be used without issues by common vcf handling tools (bcftools,svdb,igv...) and if it works well in Scout. What happen for example when variants have different svtypes or other non-numerical variables? If I understand correctly they would appear concatenated, but I don't think that makes sense and might be a problem for other tools. Or maybe I am misunderstanding how the merging works?

khurrammaqbool:
Similar to the comments in https://github.com/Clinical-Genomics/BALSAMIC/pull/1524, the overall concept of being able to identify the origin of the variants by TNScope or Vardict is important here as well. The VCF format specifications could also be preserved in a similar manner as suggested in https://github.com/Clinical-Genomics/BALSAMIC/pull/1524. Here, for the sake of clarity, I am using `combining` TNScope and Vardict variants to differentiate it from `merging` within TNScope list of variants into MNVs I will summarize it below:
1. It will be nice to see the stats that shows the effect of combining the variants from TNScope and Vardict. This method effects TGA, TO and TN cases so some results from at least one represented case with at leas following basic stats:
i . Number of SNVs and number of indels each from TNScope and Vardict before combining
ii. Number of SNVs and number of indels after combining TNScope and Vardict variants
iii. Number of combined variants
iv. The ratios of the effected variants compared to the total.
v. The Numbers and the ratio of clinically relevant variants effected by this method of combining the variants.
2. Similar to the comment in https://github.com/Clinical-Genomics/BALSAMIC/pull/1524, the interpretation relevant information will be nice to retain. This can be achieved with an approach similar to be followed in merging the variants. i.e. addition new VCF tags identified by the caller to retain the origin of the metric and keeping the defined VCF tag as unique. For the unique VCF tags, the scores should be adapted from TNScope unless inevitable as described below.
3. TNScope should take priority for the main INFO tag (AF, DP etc.) unless Vardict's output is  preferred and in that case Sentieon support should be made aware to fix the issue making it comparable/explainable about which computation to prefer  and to be able to explain the difference before deprecating Vardict. The following format can be added to the combined VCF that will help retain the maximum information for the combined variants.
```
NEW_INFO_TAG_CALLER:INFO_VAR1|ADDITIONAL_INFO_VAR1,INFO_VAR2|ADDITIONAL_INFO_VAR2; 
```
4. For the variants that are that are unique from the callers, found_in may be sufficient to identify the origin of the variants.
5. Any missing VCF tags from the other caller and the format field may make it incompatible to parse using the standard tools. A small test may be added to show that e.g. bcftools can accept the new VCF format and can perform operations on it.

BALSAMIC/assets/scripts/merge_snv_variantcallers.py
mathiasbio: This merging script merges variants only based on the criteria that they have the exact same: chrom, pos, ref, alt values, which means that unless the variant-callers are not filling the VCF format at all, they should be representing the same variant. 

Originally when I was using bcftools this wasn't the case. It seemed to be merging variants even though the alt column was different. It only looked at chrom, pos and ref. This script is more strict. 

When the variants are merged from TNscope and VarDict, the info fields from the 2 variants are merged. The logic is that unique fields from either callers will be simply added, such as NLOD and TLOD from TNscope, and "SHIFT3" from VarDict, and other fields that are shared, such as AF will be converted to lists and both values will be added. 

I have asked the Scout developers about this and they say it should be fine, as they are only reading the sample column for these metrics, and in the case of the sample column I am defaulting to taking the sample field from the 1st VCF in the list of files, which in this case is VarDict, and was chosen because of how we saw in validation of release 16 that the FLT3 variant called in both TNscope and VarDict had very different VAF in TNscope (much lower), and in general I don't feel that we're ready to let TNscope take over from VarDict yet. 

I have also uploaded VCFs processed through this script to Scout and it looked fine, but I have since refactored it a bit and I have also merged the "merge TNscope to MNV" PR into this so I want to verify this again before being entirely safe. 

I also ran this tool that you suggested in slack:  https://github.com/EBIvariation/vcf-validator on my merged MNV VCF a VCF prior to the merging with VarDict, and it showed the same error for all lines:

"Error: Format is not a colon-separated list of alphanumeric strings. This occurs 7063 time(s), first time in line 176."

But it showed the same error for the raw TNscope VCF, also the VarDict VCF shows some warnings, but not any of these errors. Regardless we haven't had any trouble in downstream processing or viewing in Scout based on these errors. So in short, the MNV post-processing isn't adding any new errors. But I will do the same for the merged VCF.

I will add some more results from testing in the PR header above, also addressing the comments from @khurrammaqbool

mathiasbio: And I realize I didn't answer your question at all! And no it takes like 5 seconds to run the script : )

fevac: very informative. Thanks for taking the time clarifying

docs/balsamic_filters.rst
```diff
@@ -326,7 +329,19 @@ An example is a MNV created by merging a phased germline SNV with a somatic SNV.
 
 - `MNV_CONFLICTING_FILTERS`: Is a filter given to MNVs with constituent variants with different filters (such as `in_normal` and `PASS`)
 
-- `MERGED_MNV`: Is a filter set to all merged MNVs, and the common filter for all constituent variants is appended after, such as `MERGED_MNV;PASS` or `MERGED_MNV;in_normal`
+.. note::
+
+    However, as we may have multiple filters which means similar things, such as germline_risk and in_normal, MNVs constituted by variants with only these filters set aren't exactly "conflicting".
```
fevac, line 334: I don't think I understand this sentence

docs/balsamic_filters.rst
```diff
@@ -326,7 +329,19 @@ An example is a MNV created by merging a phased germline SNV with a somatic SNV.
...
+
+Therefore the logic for setting `MNV_CONFLICTING_FILTERS` has been made a bit more complex, and in summary there are 3 possible outcomes for filters when merging SNVs/InDels into MNVs:
+
+1. Single filter such as PASS, when all constituting variants all have the same filter and no other.
+2. Multiple filters, such as in_normal,germline_risk, when all constituting variants have at least 1 of the matched normal filters.
+3. `MNV_CONFLICTING_FILTERS` when the merged variants have conflicting filters, and they don't all contain matched normal filters.
```
fevac, line 340: so as I understand we don't lose the information of which filters they contained but the filters information is moved to the info field, right? This in practice makes the other filters unusable, are we ok with it?

fevac (approved):
🚀

docs/balsamic_filters.rst
mathiasbio: I'll clarify! 🌟

docs/balsamic_filters.rst
mathiasbio: Basically if a MNV consists of variants that all have some form of matched normal filter set, it will have those filters kept in the merged MNV filter column. But if there is a mix of for instance 1 variant with PASS and another with in_normal, then the filter will be set to MNV_CONFLICTING_FILTERS and in a sense the normal variant will be "rescued" from potential upstream filters. However, the original filter is kept in the INFO field so it can be read in Scout. 

I also think it's ideal that these normal variants are saved in the MNV in order for VEP to make the correct protein effect prediction, a germline + somatic variant leading to a different amino acid for instance rather than if the normal is filtered out and the somatic variant is considered in isolation

Issue #1505, opened 2024-11-26
[User Story] Apply strandbias filter for WGS TN workflow


mathiasbio:
### Need

As a clinician interpreting variants from balsamic I want to avoid interpreting false positive artefacts. 

Currently in the balsamic WGS T+N workflow there's no strand bias filtering applied, which has caused a lot of likely artefacts to be called in at least a couple of cases, as can be seen in this deviation: https://github.com/Clinical-Genomics/Deviations/issues/719 where a very large number of variants were called with 100% strand bias.

These samples admittedly seems to have some technical lab issues, something to do with the index-pairs, but it highlighted a need for applying a SOR filter. 

### Suggested approach

Apply bcftools filter similar to the one applied in the tumor only WGS workflow:

`| bcftools filter --threads {threads} --include "INFO/SOR < {params.sor[0]}" --soft-filter '{params.sor[1]}' --mode '+' \`

in the rule: `bcftools_quality_filter_tnscope_tumor_normal_wgs`

### System requirements assessed

- [ ] Yes, I have reviewed the system requirements

### Risk assessment needed

- [ ] Needed
- [ ] Not needed

mathiasbio:
I have done some testing of this filter in this sheet: https://docs.google.com/spreadsheets/d/1jBSoqJR1IcSw1w5sDXeAnMSZJMrZpuxEkzW3DJeQYns/edit?gid=0#gid=0

But I'll summarise the relevant parts here.

I looked into the effects on the number of variants are applying the same filter as we're using in the WGS TO analysis on 5 different cases, where the bottom is the case from the deviation https://github.com/Clinical-Genomics/Deviations/issues/719 

If we apply the filter as it is in WGS TO there will be a LOT of variants filtered from the WGS TN workflow, stretching from between 82% -> 7.8% of the final clinical filtered variants.

I also wanted to check how the SOR parameter works for low AD variants by checking if the fraction of variants with an AD below 6 which also had the SOR > 3 filter set was higher than in the total set of variants. But I didn't see any such clear link. Which was promising.

| case   | step                   | # Variants | # SOR filtered | # below 6 AD | # SOR and below 6 AD |   | Fraction SOR to total | Fraction SOR to total below 6AD |
|--------|------------------------|------------|----------------|--------------|----------------------|---|-----------------------|---------------------------------|
[...]

PR #1506, opened 2024-11-27
feat: add SOR filter to WGS TN

mathiasbio:
## Description

As observed in deviation https://github.com/Clinical-Genomics/Deviations/issues/719 we sometimes have an issue with strandbias in the WGS TN workflow. The cause of these variants with excessive strandbias is unknown but variants with 100% strandbias as these variants had, should never have been maintained in the final clinical VCF.

Issue in balsamic: https://github.com/Clinical-Genomics/BALSAMIC/issues/1505 

#### Added

-  max SOR 4 filter to WGS TN SNV quality filter

## Documentation

- [ ] N/A
    - [balsamic_filters.rst]

### Feature Tests

- [x] Test rerun case from deviation and see that the problematic strandbias variants are removed. The shared variants between the two problematic cases are nearly entirely removed.
- [x] Investigate the change in the number of variants for at least 4 TN cases before and after this filter. In general more than half of the final variants are removed with SOR 4. 
- [x] WGS TN case is run with SOR 4

See sheet: https://docs.google.com/spreadsheets/d/1FaROC-tS8gfJp7pcC9lx-Om_nvqJ8DWfhnI5STMAG4c/edit?gid=0#gid=0

### Pipeline Integrity Tests

    - [x] N/A
    - [ ] Verified
- **TGA T/O Workflow**
    - [x] N/A
    - [ ] Verified
- **TGA T/N Workflow**
    - [x] N/A
    - [ ] Verified
- **UMI T/O Workflow**
    - [x] N/A
    - [ ] Verified
- **UMI T/N Workflow**
    - [x] N/A
    - [ ] Verified
- **WGS T/O Workflow**
    - [x] N/A
    - [ ] Verified
- **WGS T/N Workflow**
    - [ ] N/A
    - [x] Verified
- **QC Workflow**
    - [x] N/A
    - [ ] Verified
- **PON Workflow**
    - [x] N/A
    - [ ] Verified

<details>
<summary></summary>

### Documentation

- **Atlas documentation**
    - [x] N/A
    - [ ] Updated: [Link]
    - [x] N/A
    - [ ] Updated: [Link]

### User Changes

- [x] N/A

### Infrastructure Changes

    - [x] N/A
    - [ ] Updated: [Link]
    - [x] N/A
    - [ ] Updated: [Link]
    - [x] N/A
    - [ ] Updated: [Link]
- **Scout interface**
    - [x] N/A
    - [ ] Updated: [Link]

</details>

## Checklist

> [!IMPORTANT]  

### For Developers

- **PR Description**
- **Documentation**
- **Tests**
- **Review**

### For Reviewers

- **Code**
    - [ ] Code implements the intended features or fixes the reported issue.
    - [ ] Code follows the project's coding standards and style guide.
- **Documentation**
    - [ ] Pipeline changes are well-documented in the CHANGELOG and relevant documentation.
- **Tests**
    - [ ] The author provided a description of their manual testing, including consideration of edge cases and boundary
- **Review**
    - [ ] Confirmed that the developer has addressed all the comments during the code review.


fevac (commented):
Code looks good to me. I would add some info about why we are using 3 differnt SOR values to filter in the different subworkflows (2.7, 3, 4).

If possible, standardising would be best, but if not an explanation would be good

mathiasbio:
Do you mean to add it in this PR or in the docs somewhere? At the moment the reason is mostly based on variant-numbers, and ideally I would want to start from scratch with a good truthset and optimise the parameters based on this. 

But I set the 2.7 on TGA based on the twist pancancer samples that I tested earlier where no true variants were removed, and where I knew that we already had VarDict and the goal was to bring the numbers down. If we only had TNscope as a caller I would probably relax this parameter a bit, but now we were only adding more variants and I weighed a bit more towards adding more likely high quality variants (but turns out that I probably didn't filter hard enough, or at least not on the right parameter - let's hope this fixes it: https://github.com/Clinical-Genomics/BALSAMIC/pull/1526)

I don't know if SOR 3 is good for WGS TO, but it's what Gothenburg is doing in their pipeline, and what Sentieon is using in their example scripts. Then when adding an SOR filter to the WGS TN workflow in this PR, I was aware that we haven't had as many issues with number of variants in this workflow, but the only issue that we have seen was those clear strand-bias artefacts mentioned in the deviation issue. I also saw that even adding a SOR of 4 which is less stringent than the other workflows, already removed a lot of variants. So I figured it could be nice to not move too drastically, especially when the customers haven't expressed any problems with variants here. 

Basically I view it as a balancing act between precision and sensitivity, and for the workflows that don't seem to have an issue with precision I want to be more cautious in applying filters.

fevac:
Just to be clear, I am not against having different filtering settings for different workflows if there is a need for it. What I would like is to be able to trust those settings which would require documentation on why we are using these filters and trusted tests cases with clear criteria (vs. specific cases that were run at the time to take a decision but which we don't really know how they behave/what to expect). So in that sense, I fully agree with that we need a reliable truthset and revise the filtering criteria in general. So go ahead and make a more general issue, but add an specify bullet point to revise the SOR too so this doesn't fall under the radar

mathiasbio:
Added assessment issue here: https://github.com/Clinical-Genomics/BALSAMIC/issues/1540

fevac (approved):
🚀

mathiasbio:
Thanks for the review and the discussion! ❤️

